PR #62, opened 2016-07-08
Implementing SPE image

Luindil:
Hei Jerome,

i ported everything I had for reading Princeton instruments SPE images in the the fabio format. 
Most of the code is related to header information, which is also mostly used for spectroscopy, but might be useful for other users.

All individual tests are passing with:
python run_tests.py -i fabio.test.testspeimage

except for the real fabio integration, e.g. by using the fabio.open function. I guess this has to something to do with the magic numbers? The _openimage function cannot read the filetype somehow...
I hope this is easy for you to fix, because i do not completely understand why you cannot just use the file extension.

The test files are attached to this comment, because I did not know here to upload them. Also the UtilsTest.getimage functions then need to be added to the TestSpeImage class.

Best wishes,
Clemens
[spe_test_images.zip](https://github.com/kif/fabio/files/354276/spe_test_images.zip)


kif:
Hi Clemens, 

Your code looks clean: I will integrate it, put the tests in place and make the images available.
About the copyright, is it yours or the Uni-Koln ? This depends on the local law which I don't know for Germany.

Cheers,

Jerome

Luindil:
Hi Jerome,

I am not sure about this, since many things also have been done after work. However, I honestly do not care, so I think it might be better to just put: "Institute for Geology and Mineralogy, University of Cologne". --I just checked what the informatics department here is doing, and they always put the Institute also...

Thanks!

Best wishes,
Clemens

PR #65, opened 2016-07-19
FIX: QString not available on Python3.

hhslepicka:


vallsv:
Maybe it would be better to force the Qt String API.
But it is fine like that.
Thanks for the commit.

PR #73, opened 2016-10-13
Add 0 padding if the data is smaller than the expected size (close #72)

### LINKED ISSUE #72: 32bit Dectris Eiger Tif image not readable
Hi,

a Dioptas user contacted me about a tif file which he cannot open. I have the image attached. Fabio and PIL cannot handle the image. I suppose it is because of usage of 32bit UINT and not 32 bit INT.

Do you maybe have a solution for this. I think there will be more and more Eiger images, so I guess this will become a common problem.

[test.tif.zip](https://github.com/silx-kit/fabio/files/526871/test.tif.zip)

vallsv:
See issue #72


kif:
LGTM

PR #77, opened 2016-10-21
Feature fabio-convert script

vallsv:
Here is a first version for `fabio-convert`. It is not yet deeply tested. But i would like feed back.

```
$ ./bootstrap.py python scripts/fabio-convert --help
usage: fabio-convert [-h] [-V] [-v] [--debug] [-l] [-o OUTPUT] [-F FORMAT]
                     [-f] [-n] [--remove-destination] [-u] [-i] [--dry-run]
                     [IMAGE [IMAGE ...]]

FabIO-convert is a portable image converter based on FabIO library.

positional arguments:
  IMAGE                 Input file images

optional arguments:
  -h, --help            show this help message and exit
  -V, --version         output version and exit
  -v, --verbose         show information for each conversions
  --debug               show debug information

main arguments:
  -l, --list            show the list of available formats and exit
  -o OUTPUT, --output OUTPUT
                        output file or directory
  -F FORMAT, --output-format FORMAT
                        output format

optional behaviour arguments:
  -f, --force           if an existing destination file cannot be opened,
                        remove it and try again (this option is ignored when
                        the -n option is also used)
  -n, --no-clobber      do not overwrite an existing file (overrides a
                        previous -i option)
  --remove-destination  remove each existing destination file before
                        attempting to open it (contrast with --force)
  -u, --update          copy only when the SOURCE file is newer than the
                        destination file or when the destination file is
                        missing
  -i, --interactive     prompt before overwrite (overrides a previous -n
                        option)
  --dry-run             do everything except modifying the file system

return codes: 0 means a success. 1 means the conversion contains a failure, 2
means there was an error in the arguments
```
It can be used like that:
```
fabio-convert file.edf -o file.msk
fabio-convert file.edf -o directory -F msk
fabio-convert *.edf -o directory -F msk
```

It contains:
- An update of all the codec classes to provide a short description and default file extensions
- An update of utilstest to be able to test scripts
- A merge of `make_man` from `pyFAI`
- A refactoring of the import of all codecs ibnto a module `fabioformats` (`fabioutils.FILETYPES` was removed, `fabioformats` provides an API)
- the fabio-convert script with unittest, and man
- I change the visibility of `logger.error(msg)` from `FabvioImage.factory` to `logger.debug`, while an exception is already raised with the same message. It provide a better output from the script.


kif:
Hi Valentin,

There are quite a lot of typos in the help, often spaces missing. This can be related to the issue tracker of GH or to the "\n" not replaced by spaces in the source code interpretation. 

Beside this, it looks cool.

vallsv:
Nice catch. I fixed it

vallsv:
That's ready for merging

kif:
LGTM

PR #93, opened 2016-11-25
To eiger

kif:
script to convert a bunch of files into an eiger-like HDF5


vallsv:
There is a typo in ``:paran force_time``. And i think it is better to remove ``import posixpath`` instead of comment it. Else it looks fine

PR #102, opened 2017-01-30
Fix edf constructor

vallsv:
This patch add robustness on EdfImage constructor

It contains
- A refactoring to be sure the super constructor is always called
- It take care of the case where dim==0 and raise an exception
- The last patch avoid using array bigger than 2d (before that, the input data was not fully used). We can remove it if you think it is safer for compatibility.

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -474,11 +474,7 @@ class EdfImage(FabioImage):
             elif dim == 2:
                 stored_data = data
             elif dim >= 3:
-                logger.warning("Data dimension too big. Only the last 2-dimensions will be used")
-                selection = [0] * dim
-                selection[-1] = slice(None)
-                selection[-2] = slice(None)
-                stored_data = data[selection]
+                raise Exception("Data dimension too big. Only 1d or 2d arrays are supported.")
 
         FabioImage.__init__(self, stored_data, header)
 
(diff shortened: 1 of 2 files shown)


kif:
I wonder if this 3D-feature is not used by people doing tomography. Could you please check with Alessandro/Pierre ?

vallsv:
Anyway the data was cut with [0,:,:], most of the data was lost, else a single 2d image.

vallsv:
Alessandro said that they only store 2d images.

vallsv:
The Silx EdfFile implementation contains a Dim_3 and Offset_3 metadata. Then maybe it is possible to create multiframes of 3D data (kind of 4D data)? Maybe we can talk about it into another PR.

kif:
On Tue, 31 Jan 2017 00:51:06 -0800

> Anyway the data was cut with [0,:,:], most of the data was lost, else a single 2d image.

Apparently this is a bug ...

It was supposed to "auto-slice" i.e a 2,3,4-shaped array should be
represented as two frames, each 3,4-shaped. 


-- 
Jérôme Kieffer
tel +33 476 882 445

vallsv:
I see. But if it was not used like that (cause it was buggy), i think it would be better to use frame when you want to store frame. And to provide data to the constructor if you want to create an image with a single frame. That's more clear. And that's more the case if the EDF "specification" also support 3D data in a single frame.

kif:
OK.

PR #108, opened 2017-02-09
Fix OXD for pilatus300k

vallsv:
- Ascii header size is now fixed for sapphire 4.0 header (it can be a regression)
- Fix header reading for sapphire 4.0 header (it was also wrong for our testfile)
- Fix read and write of alpha/beta angles
- Add Pilatus300k detector type (detector type 5 and 6 are still missing)
- Add a new test file `pilatus300k.rod_img` provided by @Niolon


vallsv:
Jerome contacted somebody to try to have information on the specification. Maybe we can have a better patch.

kif:
LGTM

PR #110, opened 2017-03-09
Fix end of iteration for multiframe single EDF file

### LINKED ISSUE #109: Iteration over multiframe EDF raises an exception
Iterating over a 2 frames EDF files leads to ValueError exception:
```
for image in fabio.open('MultiFrame.edf.bz2'):
    print(image.currentframe)

0
1

ERROR:edfimage:Cannot access frame: 2/2
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-1530a706257b> in <module>()
----> 1 for image in fabio.open('MultiFrame.edf.bz2'):
      2     print(image.currentframe)
      3 

/users/tvincent/.local/lib/python2.7/site-packages/fabio-0.5.0a0-py2.7-linux-x86_64.egg/fabio/fabioimage.pyc in __iter__(self)
    639             yield current_image
    640             try:
--> 641                 current_image = current_image.next()
    642             except IOError:
    643                 raise StopIteration

/users/tvincent/.local/lib/python2.7/site-packages/fabio-0.5.0a0-py2.7-linux-x86_64.egg/fabio/edfimage.pyc in next(self)
    677         else:
    678             newFrameId = self.currentframe + 1
--> 679             newImage = self.getframe(newFrameId)
    680         return newImage
    681 

/users/tvincent/.local/lib/python2.7/site-packages/fabio-0.5.0a0-py2.7-linux-x86_64.egg/fabio/edfimage.pyc in getframe(self, num)
    657             txt = "Cannot access frame: %s/%s" % (num, self.nframes)
    658             logger.error(txt)
--> 659             raise ValueError("EdfImage.getframe:" + txt)
    660         return newImage
    661 

ValueError: EdfImage.getframe:Cannot access frame: 2/2
```

This seems to be due to `fabioimage.__iter__` catching `IOError` but not `ValueError` to stop the iteration.

t20100:
`edfimage.next` raises `ValueError` when reaching the end of a multiframe EDF that `fabioimage.__iter__` didn't catch.

Still, `edfimage.getframe` will log an error right before raising the `ValueError` exception.

Closes #109

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -660,9 +660,8 @@ class EdfImage(FabioImage):
             newImage.filename = self.filename
             newImage._file = self._file
         else:
-            txt = "Cannot access frame: %s/%s" % (num, self.nframes)
-            logger.error(txt)
-            raise ValueError("EdfImage.getframe:" + txt)
+            raise IOError("EdfImage.getframe: Cannot access frame: %s/%s" %
+                          (num, self.nframes))
         return newImage
 
     def previous(self):
@@ -676,7 +675,10 @@ class EdfImage(FabioImage):
         return newImage
 
     def next(self):
-        """ returns the next file in the series as a FabioImage """
+        """Returns the next file in the series as a fabioimage
+
+        @raise IOError: When there is no next file or image in the series.
+        """
         newImage = None
         if self.nframes == 1:
             newImage = FabioImage.next(self)
--- a/fabio/fabioimage.py
+++ b/fabio/fabioimage.py
@@ -284,7 +284,10 @@ class FabioImage(with_metaclass(FabioMeta, object)):
         return openimage(fabioutils.previous_filename(self.filename))
 
     def next(self):
-        """ returns the next file in the series as a fabioimage """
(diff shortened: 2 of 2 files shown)


t20100:
New proposal:
For EDF multiframe, `next` tries to go to next file in serie when reaching the last frame of current file.
There is a difference of behavior with `previous` that I haven't changed.
If you are OK with this `next` behavior, `previous` should also be changed.

Have `next` raising `IOError` if filename is not parsed as a serie filename (instead of raising `TypeError`)

vallsv:
I think `frame==1` was a test to not consider a multi frame image as part of a multi image container.

Your proposal will change the meening of it. Then I think it looks dangerous to change that.

It would be very nice to remove iteration accros sequence of images from the `Image` class and to create a dedicated class for that. Here you can iterate frames like you want.

We should talk about your usecase with Jerome.

t20100:
Yes, let's discuss that.

t20100:
Updated. Do not change the behavior of `next` except for the raised exception.

PR #168, opened 2017-09-29
166 dm3

### LINKED ISSUE #166: dm3 format under Python3
dm3 format is used in CryoEM experiments. Some sample images can be read in python2 but not in python3. It would be interesting to investigate.

Moreover the automatic detection of the format fails, i.e. fabio.open does not recognize the dm3 format while fabio.dm3image.Dm3Image.read(filename)  works in python2 but not in python3.

It is worth doing some polish on it :)

kif:
There were a couple of discrepancies: python2 vs 3 and a bug for the detection
added a test. Thanks Winnie for the image

close #166


vallsv:
```
ERROR: test_read (fabio.test.testdm3image.TestDm3Image)
check we can read dm3 images
----------------------------------------------------------------------
  File "C:\projects\fabio\build\lib.win32-3.4\fabio\test\testdm3image.py", line 79, in test_read
    obj1.read(fname)
  File "C:\projects\fabio\build\lib.win32-3.4\fabio\dm3image.py", line 124, in read
    self.read_tag_entry()
  File "C:\projects\fabio\build\lib.win32-3.4\fabio\dm3image.py", line 180, in read_tag_entry
    value = self.read_tag_type()
  File "C:\projects\fabio\build\lib.win32-3.4\fabio\dm3image.py", line 257, in read_tag_type
    swap=self.swap)[0]
TypeError: ufunc 'remainder' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
Looks to be a Python 3.3, 3.4 related problem.

vallsv:
It would be nice to clean up your commit from old ones too.

vallsv:
LGTM, but the merges of commits from august...

PR #175, opened 2017-12-01
Fix default extensions typo

vallsv:
- Fix the typo
- Update the metaclass for compatibility with the wrong typo (before deprecating it, before removing it)


kif:
LGTM. Thanks to all of you

PR #176, opened 2017-12-05
Invalidate the extension cache in case the list of formats change

### LINKED ISSUE #174: Typo in fabioimage api, "DEFAULT_EXTENTIONS"
"DEFAULT_EXTENTIONS" is misspelled. Caused a bit of confusion when I wrote "DEFAULT_EXTENSIONS" and my fabioimage subclasses didn't work; there was no indication of the source of the error, besides the class not being loaded into during fabioformats.get_all_classes().

vallsv:
Closes #174


vallsv:
Jerome and me think it would be better to use an explicit registration instead of using a metaclass.

This way would be more easy manage and to custom.

PR #180, opened 2018-01-15
Move codec registry from FabioImage class to fabioformat

### LINKED ISSUE #177: Explicit registration of file formats
Something like this:
```
class MyFormat(FabioImage):
    pass

fabio.register(MyFormat)
```
... instead of the magic done at the class construction.

It should create a more flexible library.

vallsv:
- Explicit registration of the available codecs
- Remove `registry` from `fabioimage.FabioImage` and move it to `fabioformats`
- Remove the FabioImage metaclass
- Test for the `fabio.register` function (check if adding a new format at runtime is working)
- Deprecate `FabioImage.factory`

It changes the behavior for people who are registering custom formats (now you have to call explicitly `fabio.register(MyImageCodecClass))`. Or to use it as a decorator:

```
@fabio.register
class MyImageCodecClass(FabioImage):
    pass
```

Close #177

### DIFF
--- a/fabio/__init__.py
+++ b/fabio/__init__.py
@@ -30,7 +30,7 @@ __author__ = "Jérôme Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "GPLv3+"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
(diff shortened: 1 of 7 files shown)


vallsv:
This one is also ready to merge.

PR #181, opened 2018-01-15
Backport project scripts from silx

vallsv:
- Backport setup/bootstrap/run_tests from silx
- Now it uses --installed instead of the reverse command --insource


kif:
LGTM

PR #182, opened 2018-01-15
Proposal for fabio.register function before introducing the change

vallsv:
This PR create an empty function `fabio.register` for the release of fabio 0.6. It allow people to patch there code before the change the coming change of the behaviour in fabio 0.7.

That's just a proposal in case we do not apply the change now.


kif:
Two comments:

* I would like this register function to be useable as a decorator, so it should return the input class.
* As it does nothing for now, is should print a warning saying it is a dummy function which does not alter the behaviour of the function

Beside this it is OK for v0.6

PR #185, opened 2018-01-16
Improve jpg2k support

vallsv:
- Clean up log
- Robustness of PIL and glymur versions on Debian 7


vallsv:
It should be merged. Looks better than before.
Else at least we should remove the `print` :-)

PR #194, opened 2018-01-24
add magic number for  Bruker100 images

t-onoz:
This will facilitate opening New Bruker (bruker100) format frames.


vallsv:
Thanks a lot for the patch. But no idea why it was not already there.
Could you please send us a sample in this format for our regression tests. Something we can use with the MIT license.

t-onoz:
Here is a sample file with Bruker100 format.
https://www.dropbox.com/s/lbvmt5uwm9c6q3x/bruker100_v500.gfrm?dl=0

PR #201, opened 2018-02-02
Improve pilatus tiff support

### LINKED ISSUE #202: TIFF headers from TiffIO contains \x00

vallsv:
- Rework the way tiff and pilatus are reading headers
- Now Pilatus metadata are exposed instead of the TIFF metadata
- Support metadata when image is loaded by PIL
- Support multiframes
- `image.close()` should be called at the end of the use
- TAG_MODEL is now exposed (could be used for further patches, cause it contains the fact it was generated by a Pilatus)
- Rework the way ASCII fields are parsed from TiffIO (fields are exposed as text, not as binary strings) Closes #202. We can talk about this change, another way is to only patch the `info` dict in this way.
- Test a Pilatus1M TIFF file provided by @mikapfl


vallsv:
@mikapfl here is an update of the patch. You are credited for a commit, hope the author field is fine for you.

There is much more changes than expected, will see if it is fine.

kif:
That's indeed a big chunk !

PR #205, opened 2018-02-15
TiffIO: Allow custom description and software tag

diorcety:
Allow tiff image without software tag and a description which is not a dict

### DIFF
--- a/fabio/TiffIO.py
+++ b/fabio/TiffIO.py
@@ -676,7 +676,7 @@ class TiffIO(object):
                 readout = numpy.fromstring(fd.read(nBytes), dtype)
             if hasattr(nBits, 'index'):
                 readout.shape = -1, nColumns, len(nBits)
-            elif info['colormap'] is not None:
+            elif colormap is not None:
                 readout = colormap[readout]
             else:
                 readout.shape = -1, nColumns
@@ -727,7 +727,7 @@ class TiffIO(object):
                         readout = numpy.fromstring(bufferBytes, dtype)
                     if hasattr(nBits, 'index'):
                         readout.shape = -1, nColumns, len(nBits)
-                    elif info['colormap'] is not None:
+                    elif colormap is not None:
                         readout = colormap[readout]
                         readout.shape = -1, nColumns, 3
                     else:
@@ -1244,9 +1244,9 @@ if __name__ == "__main__":
     for i in range(tif.getNumberOfImages()):
         info = tif.getInfo(i)
         for key in info:
-            if key not in ["colormap"]:
+            if key != 'colormap':
                 print("%s = %s" % (key, info[key]))
-            elif info['colormap'] is not None:
+            elif info[key] is not None:
                 print("RED   %s = %s" % (key, info[key][0:10, 0]))
                 print("GREEN %s = %s" % (key, info[key][0:10, 1]))
                 print("BLUE  %s = %s" % (key, info[key][0:10, 2]))
--- a/fabio/TiffIO.py
+++ b/fabio/TiffIO.py
@@ -786,12 +786,7 @@ class TiffIO(object):
 
         return image
 
-    def writeImage(self, image0, info=None, software=None, date=None):
-        if software is None:
-            software = 'PyMca.TiffIO'
-        # if date is None:
-        #    date = time.ctime()
-
+    def writeImage(self, image0, info=None, software='PyMca.TiffIO', date=None):
         self.__makeSureFileIsOpen()
         fd = self.fd
         # prior to do anything, perform some tests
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM
Thank for the PR

diorcety:
Sorry I maybe miss some other stuff. Why using Unknown software or Unknown date instead of not putting the entry in the info directory (read part)?

kif:
On Thu, 15 Feb 2018 16:29:48 +0000 (UTC)
Diorcet Yann <notifications@github.com> wrote:

> Sorry I maybe miss some other stuff. Why using Unknown software or Unknown date instead of not putting the entry in the info directory (read part)?

Hi Yann,

Thanks for your contribution, FabIO has a long history (>10years) and
while TiffIO has other roots, it is not very recent code. @vasole is
probably the one who can answer this.

Tell me if you want to visit us.

Cheers,

Jérôme

vasole:
- The strange "%s" % "" dates from the early days of Python 3
- The reason to require a dictionary as info is to allow to recover a dictionary in the same way as other formats like EDF. If I am not mistaken, ImageJ was able to understand having key values too. 
- If when reading one cannot get the field TAG providing the name of the software that wrote the image, it seems appropriate to read "Unknown Software". Any other thing would be either lying or saying something with little probability of being true.
- If when reading one cannot get the field TAG providing the date the image was written , it seems appropriate to read "Unknown Date".  Any other thing would be either lying or saying something with little probability of being true.

vasole:
As far as I can understand from the interpretation of the TIFF format, that information has to be there. If not present, then unknown software and unkown date seem fine to me.

See page 20 in:

https://www.itu.int/itudoc/itu-t/com16/tiff-fx/docs/tiff6.pdf

vasole:
To allow for an info that is not a dict is fine **as far as it can be converted to a bytes object** (reason of the printing stuff for which today there are better alternatives).

vasole:
In short:

- I would not allow to read or write an image without the software tag.
- I would allow to pass an info that is not a dict provided it is made writing ``description = "%s" % info`` and not ``description = info`` as proposed.

vallsv:
If there is no software tag, i don't see the point to display anything. The fact there is no software tag is an information by itself, and the way you display it can be done in many different ways. Then i can't agree.

diorcety:
I'm agree with @vallsv, I fixed the PR

vasole:
You should look a bit further away. 

The software field **has to be there**. If you want to let it as an empty string instead of "Unknown Software" it is your business but the field has to be there. Why? Easy. As you can see you can write anything as description (particularly after your pull request). Therefore to interpret that description one needs that field.

vallsv:
We still have to talk on our side, cause there is no consensus. But at least we agree with the idea of the change on the `info` on the `write`.

For the tags like `software` and `date`, it looks to be mandatory fields of the TIFF format, that is why Armando prefer to expose something. Do you have problem when you read your own files? or when you read third party files? Did you try to write your files with `info={'software': 'MySoftware', 'date': '2018'}`?

diorcety:
> For the tags like software and date, it looks to be mandatory fields of the TIFF format
I'm not totally agree with that. I updated the PR according to the fact that date and software have as default value None. So the fields are still here now. Just using the default value from the spec

diorcety:
The issue is that if I open a file with fabio, change one pixel, save the file, the saved file is far different from the original one

vasole:
Is your input file TIFF? Have you tried other TIFF writer modules? (TIFF = Thousands of Incompatible File Formats)

 TiffIO.py  was mainly thought for reading uncompressed and packbits compressed TIIF files while offering the same writing functionalities as EdfFile.py and that achieves it. If you try to use it as a generic TIFF converter you are using the wrong tool.

vasole:
Personally I would expect an empty string and not None because:

- I doubt the specs accept to return an object where a string is expected
- Returning None where a string was returned can certainly break code

vasole:
I'm not directly responsible of this project and my main concern is if once we merge our three different implementations. From my side I would accept the pull request if

- There is a conversion of the user supplied info (not necessarily a dict) to a string ``description = "%s" % info`` or an assert is provided to avoid writing corrupted files because for instance an array or a list is passed as info.
- The nature of the returned software and date fields is not changed. I know it is more pythonic to return None than an empty string but a) one expects a string and b) it can break software

vallsv:
I was afraid you was using something like `from fabio import TiffIO`. AFAIK this module should be private then if you use it through the `FabioImage` class it's fine.

I will wait for Jerome @kif before accepting this PR, with some changes. I guess with empty string instead of `None` as requested by Armando to avoid to break other users. I will also move this module as private to avoid to import it, as fabio is not considered to provide the reference of `TiffIO`.

As said by Armando, If you want to edit a file an still have the header unchanged, i think you should check for another library. You can check this list of library shared by Gohlke on his library: https://pypi.python.org/pypi/tifffile

BTW, which kind of detector do you use?

diorcety:
The issue is not only for editing one tiff image. If I convert one tiff to edf, some header are added like software and date.

kif:
Vallentin has fixed the 2 issues in a couple of other PR

PR #207, opened 2018-02-26
EDF: clear capsHeader before update it

diorcety:
If you (only) remove some reserved header (ex DIM_1) from a edf image, when you write this image an exception is raised: KeyError: u'Dim_1'

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -366,6 +366,8 @@ class Frame(object):
         else:
             data = self.data
         fit2dMode = bool(fit2dMode)
+
+        self.capsHeader.clear()
         for key in self.header:
             KEY = key.upper()
             if KEY not in self.capsHeader:


vallsv:
Hi. We don't have much time this week to look at it.

But could you please send us your use case. I don't think you are supposed to change the `capsHeader`, that is why i ask. If you could provide few lines of code to reproduce the problem, it will really useful.

Thanks a lot.

diorcety:
Example

> img = read_edf("/tmp/abcd")
> del img.header['Dim_1']
> img.write("/tmp/abcd2")

Due to the fact that capsHeader is set in read_edf, if you remove reserved key from header this will produce an exception (even if this header will be computed again)

vallsv:
Makes sens. I will ask Jerome what to do. At least it have to work.
We also think about storing reserved EDF headers into another python header. But as ever, it can break things on user side.

vallsv:
I will rework your PR, but the main idea is fine.

vallsv:
I merge as it is and will rework from the master, as everything is fine on the CI.

PR #209, opened 2018-03-02
Clean up capsheader from EDF

vallsv:
- Remove self.capsHeadcer
- Create and use it only locally when reading/writing EDF headers
- Remove the `capsHeader` API


kif:
LGTM

PR #214, opened 2018-03-26
Fix rst README for PyPI

takluyver:
The [package page on PyPI](https://pypi.python.org/pypi/fabio) currently shows the raw rst markup, which looks especially ugly when it's not in a monospace font. When the rst is valid, it shows as nicely rendered HTML instead.

I used the [readme package](https://pypi.python.org/pypi/readme) to find the changes needed.


vallsv:
LGTM, thanks a lot for the fix

PR #223, opened 2018-06-11
Rework appveyor config file using silx template

vallsv:
Some scripts are outdated and do not work anymore, this PR try to use an up-to-date configuration for python installation.


t20100:
You might want to consider using `--no-cache-dir` and `--no-index` `pip install` options.
In `silx` we use: `- "pip install --pre --find-links dist/ --no-cache-dir --no-index silx"` to make sure we don't install from pip cache or from pypi, I think we had an issue with that at some point.

vallsv:
Thanks for the trick, i will do it into another PR.

PR #224, opened 2018-06-11
Support pathlib

### LINKED ISSUE #215: Allow PosixPath type for filenames.
It would be useful to support pathlib Path objects (i.e., PosixPath) for filenames.

For Python 3 just changing:
### fabioutils.py
Changing:
```
StringTypes = (str, bytes)
```
to:
```
from pathlib import PosixPath
StringTypes = (str, bytes, PosixPath)
```
was enough in my very quick test. Not sure what would be involved for Python 2.

I can try creating a Pull Request if this is seen as useful.

vallsv:
Here is a path to support `pathlib`.

I am not very happy with it code the architecture do not have a good file abstraction, with a lot of API, then there is a lot of duplication. But it is working.

Let me know if it is useful.

The "Clean up tests" commit is a random clean up which have no relation with this PR. It can be part of another PR if you prefer.

Closes #215


kif:
LGTM

PR #227, opened 2018-06-12
Improve CI scripts

vallsv:
- Fix `--no-cache-dir --no-index` on pip
- Only test the project 1 time on appveyor (it's the same as travis), instead of 4 times
- Backport grouping skipped test reason from silx


kif:
LGTM. let's wait fro appveyor

PR #231, opened 2018-06-13
Create a file series object based on FabioImage

vallsv:
This PR feature a file series object based on `FabioImage`.

It is designed to do a full abstraction of a file series. As result the set of the file is considered as a single container of the whole frames contained in this files.

- It features a random and sequential access to frames
- Allows to know the number of frames
- Random access can be optimized if we know the amount of frames per files.
- Plus in case of EDF multi frames the @t20100 lazy iterator reduce by 2 to reading time. And only load a single frame at a time (could reduce memory access too).

```
# Hi-level API
import fabio
series = fabio.open_series(first_filename="foo0000.edf")
# or
series = fabio.open_series(filenames=["foo0000.edf", "foo0001.edf"... ])
# or
series = fabio.open_series(filenames=["foo0000.edf", "foo1000.tif", "mydata.h5"])
```

```
# Can be optimized if we know the way files are stored
# In this case `get_frame` and `nframes` access are optimized
series = fabio.open_series(filenames=filenames, single_frame=True)
series = fabio.open_series(filenames=filenames, fixed_frames=True)
series = fabio.open_series(filenames=filenames, fixed_frame_number=100)
```

```
# Random access to frames
frame = series.get_frame(200)
frame = series.get_frame(201)
frame = series.get_frame(10)
frame.data
frame.header
```

```
# Iterate through all the frames
for frame in series.frames():
    frame.data
    frame.header
```

```
# Close opened files
series.close()

# or use with
with fabio.open_series(first_filename="toto.edf") as series:
    pass
```


kif:
LGTM

PR #233, opened 2018-06-19
Fix TIFF image using RGB LUT

### LINKED ISSUE #232: pb opening tiff file using Fabio
I rely on Fabio for file input but I regularly run into problems when reading tiff files. I attach an example below.
Q1: I don't really understand what the pb is here, the size does match...
Q2: it seems that PIL somehow modifies the output (the statistics of the loaded array doesn't match with this of the image loaded with ImageJ). What is done exactly when falling back onto PIL ? I use pillow 5.0.0

The test file can be downloaded from: https://www.dropbox.com/s/jcbd73shy7aascv/_0001_C1_DEJ_equal_0060.tif?dl=0

---
#using v 0.6.0 under python 2.7.14
import fabio

fname = 'D:/Temp/_0001_C1_DEJ_equal_0060.tif'

dataarr = fabio.openimage.openimage(fname).data

>>WARNING:fabio.TiffIO:Bogus StripByteCounts information
WARNING:fabio.tifimage:Unable to read h:\Dentin\ZoneModele_0001_C1_DEJ\ZoneModele_0001_C1_DEJ\Seuillage\MyTests\_0001_C1_DEJ_equal\_0001_C1_DEJ_equal_0060.tif with TiffIO due to could not broadcast input array from shape (17304000,3) into shape (8400,2060,3), trying PIL

vallsv:
- Fix TiffIO (the patch is not ready, check it with @vasole)
- Fix PIL fallback to use the `P` mode (palette), the general case is i guess more complex, but it is already fine.

The result is a 3 dim image (width, height, channel)

Closes #232

### DIFF
--- a/fabio/TiffIO.py
+++ b/fabio/TiffIO.py
@@ -684,9 +684,10 @@ class TiffIO(object):
                 readout.shape = -1, nColumns, len(nBits)
             elif info['colormap'] is not None:
                 readout = colormap[readout]
+                readout.shape = -1, nColumns, 3
             else:
                 readout.shape = -1, nColumns
-            image[rowMin:rowMax + 1, :] = readout
+            image[rowMin:rowMax + 1] = readout
         else:
             for i in range(len(stripOffsets)):
                 # the amount of rows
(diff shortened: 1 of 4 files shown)


vallsv:
@vasole I created 2 patchs:

- The one we discussed cfca350
- And another one 7de81c6 (cause `nBytes` was wrong with TIFF i created using Gimp)

I also suggest to merge the 2 branches from `if len(stripOffsets) == 1:`, the code should be the same for both cases, but it is not (the code for the colormap was already fine on the `else` branch for example).

vallsv:
It is the case in this `if` branch, that is why i propose a merge.

PR #235, opened 2018-06-25
Use numpy.frombuffer

kif:
Copy data only if they may be modified
byteswap them in place where-ever possible
Remove commented-out code (if 0)


vallsv:
LGTM

PR #239, opened 2018-06-26
Remove URL from openimage

### LINKED ISSUE #238: eigerimage on windows and URLs

vallsv:
- URL is never used
- Fix regression with absolute Windows path using slash instead of backslash + dataset path `"C:/something.h5::group/entry"`

Work around:
- Use relative path

Closes  #238


kif:
LGTM

PR #254, opened 2018-09-14
EDF multi-frame iterator

t20100:
This PR proposes to add a `fabio.edfimage.edf_lazy_iterator` function that returns an iterator over the frames of a single multi-frame EDF file.

The benefit is to avoid reading the whole file to get the number of frames and build the index before doing anything.

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -1042,3 +1042,72 @@ class EdfImage(FabioImage):
 
 
 edfimage = EdfImage
+
+
+def edf_lazy_iterator(fname):
+    """Iterates over the frames of an EDF multi-frame file.
+
+    This function optimizes sequential access to multi-frame EDF files,
+    by avoiding to read the whole file at first in order to get the number
+    of frames and an index of frames for faster random access.
+
+    Usage:
+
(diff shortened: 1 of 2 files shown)


vallsv:
On #231 i created a file serie iterator based on FabioImage. It uses a new `frames` API which provides  the iterator. It could be good to create the same for EddImage, instead of creating a function API as you do.I think it is quite explicit and easy to use.

As result:
```
image = fabio.open("foo.edf")
for f in image.frames():
    pass
```

vallsv:
The point is I don't really like the idea of breaking the abstraction, and creating a side API. But it's up to you.

Yes it is right. It could be managed by the class if the loading was lazy. This also could be customed by a constructor parameter.

Is your PR urgent? UI could try to merge it in this way if you like the idea. `frames` is anyway not yet part of the API then...

t20100:
I also think it is best to have a single API (and a single implementation because I did quite a bit of copy-paste here)...
And if possible without intervention from the user... which need to support lazy-loading of `nframes` and eventually populate the multiframe index lazily...

For urgent need, the code of this PR can be copied to a separate file... So if there's a chance to merge it in the standard API, it is better.

kif:
I agree with Thomas

PR #256, opened 2018-10-10
fix debian autopackaging ...

kif:
There are 2 bugs: devscripts providing sch is not required and there was an unconsistancy between fabio and python-fabio. both are fixed now


vallsv:
LGTM

PR #263, opened 2018-10-23
Fix EDF compression warning

vallsv:
Clean up the EDF compression parsing

- Fix test (warn only if there is NO compression)
- Log as warning, not info

@diorcety i reuse your patch on EDF file reader. Let me know if you still have problems


kif:
LGTM

PR #265, opened 2018-10-23
Merge EDF frame iterator

### LINKED ISSUE #258: [edf] Merge the EDF frame iterator as the default EdfImage API
Merge `edf_lazy_iterator` inside `EdfImage`

See #254

vallsv:
Here is a merge to provide a lazy loadable file image.

- It speed up reading big multi-frames
- But there is no more `IOError` execptions when the file object is created

Then it could be nice to:
- [ ] Create a `sync()` function to be sure everything was read.
- [ ] Or add a flag to the constructor to enable the feature
- [ ] Be able to call `nbframes` inside the `frames()` loop (it's tricky)

```
image = fabio.open("file.edf")
for frame in image.frames():
    pass
```

Closes #258

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -914,6 +914,22 @@ class EdfImage(FabioImage):
     # Properties definition for header, data, header_keys
     ############################################################################
 
+    def _get_any_frame(self):
+        """Returns the current if available, else create and return a new empty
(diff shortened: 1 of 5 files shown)


vallsv:
I close it. That's not the way to proceed anyway.

PR #273, opened 2018-10-29
Small clean up

vallsv:
- Typo to prepare a shared class for frames
- Test PiXi format
- Clean up PiXi module
- Test TiffIO module
- Remove useless comments
- Remove `__name__ == "main"`
- `Hdf5Image` now returns an `Hdf5Frame` instead of an `Hdf5Image` (it could be a regression in case people uses `isinstance(f, Hdf5Image)`. I don't think it is a blocking patch)


kif:
LGTM

PR #275, opened 2018-11-09
fixes to dm3image

ebknudsen:
A report (and suggested fix) came in from G. Mangan, about dm3image failing when only a single 'Binning' item was present in the file. This fix allows either one or two numbers. Also a small cleanup.


vallsv:
Thanks a lot for the patch. `eval` and `go_on` was really dirty.

vallsv:
BTW if you like, if you have a small dm3 with a single binning value, we could use it as resource for our regression tests.

PR #276, opened 2018-11-12
Big rework to provide frames and shape/dtype

vallsv:
This PR provides mostly 2 features:
- Add concept of frames
- Uses `shape` and `dtype` (a-la `numpy` or `h5py`)

But will also patch things like:
- A lot of the API will be deprecated (it will be displayed on debug only for this version)
- It could break third party codec
- Part of the current `EdfImage` API will not exists anymore
- Getter compatibility will still exist (`bpp`, `bytecode`, `dim1`, `dim2`, `dim3`)


kif:
Let's give it a try ... and check how other code behaves

PR #277, opened 2018-11-15
Deprecation for 0.10.0

vallsv:
This PR plan the deprecation warnings for fabio 0.10.0beta, while allowing to display pending deprecation warning if requested.

```
# Here nothing is displayed
import fabio
image = fabio.open("../data/edf/medipix.edf")
image.getData()
```

```
# Using this logging command will allow project to check and fix warnings using fabio 0.9
import logging
logging.getLogger("fabio.DEPRECATION").setLevel(logging.DEBUG)

import fabio
image = fabio.open("../data/edf/medipix.edf")
image.getData()
DEBUG:fabio.DEPRECATION:Function getData is deprecated. Reason: Prefer using 'img.data'.
  File "<ipython-input-5-94eb0a625094>", line 1, in <module>
    image.getData()
```

### DIFF
--- a/fabio/utils/deprecation.py
+++ b/fabio/utils/deprecation.py
@@ -1,0 +1,119 @@
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2016-2017 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


fabio/utils/deprecation.py
```diff
@@ -0,0 +1,147 @@
...
+    :param Union[int,str] deprecated_since: If provided, log the deprecation
+        as warning since a version of the library, else log it as debug.
+    """
+    if not depreclog.isEnabledFor(logging.WARNING):
+        # Avoid computation when it is not logged
+        return
+
+    msg = "%s %s is deprecated"
+    if since_version is not None:
+        msg += " since silx version %s" % since_version
```
kif: Are you sure it is silx ?

fabio/utils/deprecation.py
```diff
@@ -0,0 +1,147 @@
...
+               only_once=True, skip_backtrace_count=1,
+               deprecated_since=None):
+    """
+    Decorator that deprecates the use of a function
+
+    :param str reason: Reason for deprecating this function
+        (e.g. "feature no longer provided",
+    :param str replacement: Name of replacement function (if the reason for
+        deprecating was to rename the function)
+    :param str since_version: First *silx* version for which the function was
```
kif: FabIO ?

version.py
```diff
@@ -93,11 +92,48 @@
...
+    """Calculate the hexadecimal version number from the tuple version_info:
+
+    :param major: integer
+    :param minor: integer
+    :param micro: integer
+    :param relev: integer or string
+    :param serial: integer
+    :return: integer always increasing with revision numbers
+    """
+    if string is not None:
```
kif: I wonder if 
try:
    parts = string.split(".", 2)
    parts[2], releaselevel = parts[2].split("-",1)
    major, minor, micro = (int(i) for i in parts)
except Exception as err:
    raise ValueError("'%s' is not a valid version" % string)

It is twice shorter (6 vs 12 lines), prevents the use of one global, one extra module, especially THE one responsible for so many regression in python3.7 (not to mention I don't  like regular expression)

vallsv: Your proposal do not match the version format as it is provided by fabio. Then i don't think it's a good idea.

I personally have no problems with the amount of lines, the extra standard module, nor the global as we use it everywhere. But we could move this function in fabioutils, which is IMO a better place, or uses to pkgsrc (an extra dependency) parser which was designed for this usecase.

fabio/utils/deprecation.py
vallsv: Oups

PR #280, opened 2018-11-29
build on debian10/buster

kif:
packaging fro debian10


kif:
tested on scisoft10/11 freshly re-installed

PR #283, opened 2019-01-08
Fix python 3.3

vallsv:
lxml do not support anymore python 3.3


kif:
nice way to maintain the compatibility with python 3.3 ...
may be drop one of those days (probably with python2.7)

PR #291, opened 2019-03-01
Close HDF5 from fabio image

vallsv:
HDF5 file was not closed when the FabioImage was not anymore needed.

This should improve the things especially on Windows.


kif:
indeed, it is important to close files

PR #296, opened 2019-03-07
Backport 0.9 to master

vallsv:


kif:
I guess there is more to do before 0.9 final

PR #298, opened 2019-03-07
Fix from beta feedback

### LINKED ISSUE #297: Problem to open a floating-point  EDF image
Feedback from Peter Boesecke

```
In [66]: img=fabio.open("sphere.edf")
In [67]: data=img.data
/mntdirect/_users/boesecke/programs/work/silx/devel-peter/lib/python3.5/site-packages/fabio/edfimage.py in _unpack(self)
    394                 logger.info("Data stream contains trailing junk : %s > expected %s bytes" % (obtained, expected))
    395                 rawData = rawData[:expected]
--> 396             data = numpy.frombuffer(rawData, self._dtype).copy().reshape(shape)
    397             if self.swap_needed():
    398                 data.byteswap(True) 
```

From Peter Boesecke:

> The size of the array is Dim_1=400, Dim_2=300, DataType = FloatValue => 400 * 300 * 4=480000 bytes
> 
> But the size of the binary block is padded to a block boundary of 512 bytes: => 480256 bytes
> 
> As in the case of compressed data, you must not use the edf Size header parameter for cross checking the array shape! The Size parameter tells you only that you cannot read more than 480256 bytes from the file!
> 
> You must read 400 * 300 * 4 bytes = 480000 bytes. There will not be any problem for reading any historic edf file after correcting it! The wrong assumption that Size is equal to the array size dates back to the fortran routines used in onze. Inside these routines Size bytes are read into an array with only Dim_1 * Dim_2 * DataSize bytes allocated. Of cause, therefore it crashes regularly.

vallsv:
- Fix setter to dim1, dim2 (`set_dim1` was existing, then i used it to patch the `data.shape`)
- Fix setter to `pilimage = None`
- Fix `open_series` documentation
- Use the provided temp_dir instead of using a new one
- Move codec tests into `fabio/test/codecs`
- Merge `ExternalResources` from silx
- Normalize `UtilsTest` as it is provided in silx
- Fix warnings relative to small blocks
- Take a look at the float EDF from (https://github.com/silx-kit/fabio/issues/297)

Closes #297


kif:
From what you have done, It looks like *silx* did it the proper way.

kif:
Great !

PR #300, opened 2019-03-15
Improve adsc format

vallsv:
Here is small typo to clean up to provide better debug message, and a better ADSC reader.


kif:
Looks good

PR #301, opened 2019-03-15
Improvement to read d*TREK

vallsv:
This is a dirty patch to read a very specific d*TREK format using the ADSC reader.

The data read was not checked, then maybe it is still not working anyway.

Only the last commit is dirty, the other ones are part of the PR #300

The parsing of the d*TREK should be done in a specific reader and not merged into the ADSC one. But maybe what we call ADSC in fabio is d*TREK, or maybe d*TREK was designed as a superset of ADSC. No idea.


vallsv:
A better PR can be done

PR #303, opened 2019-03-15
Implement d*TREK instead of ADSC

### LINKED ISSUE #299: Rigaku's new Hypix detector
I would like to request that this image type be added.  I have attached an image for your examination.  The format of this image is d*TREK with an ACSII header.  The files can be read using ADXV, but neither RAW nor fabio can open these files.  Here is more information for you:

```
HEADER_BYTES= 5632;
BYTE_ORDER=little_endian;
PXD_DETECTOR_ACQUISITION_BIT_SHIFT=31 0;
PXD_DETECTOR_ACQUISITION_DATAMODE=Combined 31bit;
PXD_DETECTOR_DESCRIPTION=HyPix-3000;
PXD_DETECTOR_DIMENSIONS=775 385;
PXD_UNBINNED_BEAM_POSITION=677.6 190.5;
```

If you need more information, I would suggest contacting:  Angela.Criswell@rigaku.edu
Thanks,
Steve
[HSA_1_5mg_C1_0004.img.gz](https://github.com/silx-kit/fabio/files/2967687/HSA_1_5mg_C1_0004.img.gz)

vallsv:
This PR implement the d*TREK specification instead of the ADSC we was used.

The specification is provided as a PDF which make it easy to implement.

If you expect to write ADSC format the way it was done previously, this PR will maybe break your file. If you use fabio to read and write your files, it's anyway fine.

- Rename module and class from `adscimage` to `dtrekimage` (references are still there in case)
- Allow to read/write `int`, `uint` (8, 16, 32) + float32
- Previously only uint16 was supported
- Write all mandatory header metadata (except header version, it looks to be non mandatory)
- Rework the reader as a steam (remove seek -> no more problems with gzip)
- Add unittest for the writer
- Add sample from HyPix-3000 (thanks to @stevestayrook)

Closes #299


kif:
looks good!

PR #311, opened 2019-04-23
Add ehf

boesecke:
Adds ehf-type edffile reader: 
- adding magics for extended edf-files (ehf-type) 
- reading a general block
- reading binary data from external files (gzipped or not)
- distinguishing image.psd frames image.error (variances) frames

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -333,6 +333,18 @@ def get_frame_idx( frames ):
 
     return frame_idx
 
+def get_block_id_numbers( frames ):
+    '''
+    This function extracts from the list of frames the list of block_id_numbers.
+    img._block_id_numbers[:img._frame_idx] returns the list of psd block ids
+    img._block_id_numbers[img._frame_idx:] returns the list of error block ids
+    '''
+    block_id_numbers = []
+    for nframe in range(len(frames)):
+        block_id_numbers.append(frames[nframe]._block_id_number)
+
+    return block_id_numbers
+
 class MalformedHeaderError(IOError):
     """Raised when a header is malformed"""
     pass
@@ -884,6 +896,8 @@ class EdfImage(fabioimage.FabioImage):
         #      nframes=10, frame_id = 4 => errorframe_id=5
         self._frame_idx = get_frame_idx( self._frames )
 
+        self._block_id_numbers = get_block_id_numbers( self._frames )
+
         # generalframe
         self.generalframe = generalframe
 
@@ -1136,6 +1150,7 @@ class EdfImage(fabioimage.FabioImage):
         self._frames = []
         self.generalframe = None
         self._frame_idx = 0
+        self._block_id_numbers = []
 
         header_size = None
         binary_size = None
@@ -1199,10 +1214,12 @@ class EdfImage(fabioimage.FabioImage):
             elif frame._chain_number>0:
                 # add a psd frame
                 self._frames.insert(self._frame_idx,frame)
+                self._block_id_numbers.insert(self._frame_idx,frame._block_id_number)
                 self._frame_idx+=1
             else:
                 # add an error frame
                 self._frames.insert(self._frame_idx,frame)
+                self._block_id_numbers.insert(self._frame_idx,frame._block_id_number)
 
             if frame._chain_number!=0:
                 # Check the header information, because it is a standard block
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -122,7 +122,7 @@ MINIMUM_KEYS = set(['HEADERID',
 
 MINIMUM_KEYS2 = set([ 'EDF_DATABLOCKID', # Replaces HeaderID and distinguishes Error and Image frames
                       'EDF_BINARYSIZE',  # Replaces Size
-                      'BYTEORDER', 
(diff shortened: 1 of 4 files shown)


vallsv:
I think it would be better to use one of this API instead of mixing together psd/error frames.

```
        image = frame=fopen(filename)
        for frame in image.frames():
            variance = image.get_variant(frame.frame_id)
            variance.data
            variance.header

        image = frame=fopen(filename)
        for frame in image.frames():
            frame.variance.data
            frame.variance.header

        image = frame=fopen(filename)
        for frame in image.frames():
            frame.variance_data
            frame.variance_header
```

kif:
On Tue, 23 Apr 2019 06:28:59 -0700
pb38k <notifications@github.com> wrote:

> Adds ehf-type edffile reader: 
> - adding magics for extended edf-files (ehf-type) 
> - reading a general block
> - reading binary data from external files (gzipped or not)
> - distinguishing image.psd frames image.error (variances) frames

From a metric point of view, your PR is 900 loc while edfimage was
1200, making this PR huge. You should consider splitting it into multiple
parts.

Beside this, I have the feeling this EHF format differs from the EDF
format substantially (external data, possibly compressed). Wouln't it
make sense to have an ehfimage file which builds on top of edfimage ? 
It would make your modification less impacting for other users.

Is the distinction between PSD and ERROR (variance ?) part of EDF or
part of EHF ? I did not clearly see the difference.

The tests provided are not passing due to the absence of a "suite" function
building the test-suite for your ehfimage. Not that is "important" but
most reviewer look at the code only when all lights are green.


Cheers,
-- 
Jérôme Kieffer

vallsv:
My bad for the test suite.

boesecke:
> 
> 
> I think it would be better to use one of this API instead of mixing together psd/error frames.
> 
> ```
>         image = frame=fopen(filename)
>         for frame in image.frames():
>             variance = image.get_variance(frame.frame_id)
>             variance.data
>             variance.header
> 
>         image = frame=fopen(filename)
>         for frame in image.frames():
>             frame.variance.data
>             frame.variance.header
> 
>         image = frame=fopen(filename)
>         for frame in image.frames():
>             frame.variance_data
>             frame.variance_header
> ```

Your suggestion is quite obvious and I have therefore looked for ways of implementing it without changing the basic structure of the edf-reader. So far, I did not find a simple way 
without changing  important parts of the reader! I need to parse first through the whole file first! All frames (primary data and variance frames) must be available in the frames list of the EdfImage class. Currently, variance and primary data frames are read in the same way. There is only a trial done depending on the first keyword in the header. Standard frames are located at the beginning of the internal frame list, variance frames are located at the end. But all frames, whether they are primary data frames or variance frames are still frames as before and can be processed by fabio without any further modification.
Therefore there is also no problem of reading older files. They only contain primary data frames and are accessible in the historical way. I do not see any problem reading old files.  At the opposite, I have found some hidden problems that should now be resolved, e.g. if the header end pattern is crossing a block boundary.
Implementing your suggestion is a bigger work and could create unnecessary problems. 
As alternative I like creating two lists in the EdfImage class, one with the frame numbers for normal data frames and one with the frame numbers of variance data frames . These lists can alternatively be used, when needed.

boesecke:
> 
> On Tue, 23 Apr 2019 06:28:59 -0700 pb38k ***@***.***> wrote: Adds ehf-type edffile reader: - adding magics for extended edf-files (ehf-type) - reading a general block - reading binary data from external files (gzipped or not) - distinguishing image.psd frames image.error (variances) frames
> From a metric point of view, your PR is 900 loc while edfimage was 1200, making this PR huge. You should consider splitting it into multiple parts. Beside this, I have the feeling this EHF format differs from the EDF format substantially (external data, possibly compressed). Wouln't it make sense to have an ehfimage file which builds on top of edfimage ? It would make your modification less impacting for other users. Is the distinction between PSD and ERROR (variance ?) part of EDF or part of EHF ? I did not clearly see the difference. The tests provided are not passing due to the absence of a "suite" function building the test-suite for your ehfimage. Not that is "important" but most reviewer look at the code only when all lights are green. Cheers, -- Jérôme Kieffer

Dear Jerome, after adding the missing function all unittests have passed. 
I will push the new version to the server, to get green lights, I hope.

kif:
Hi Peter,
If I understand well, there are 2 topics, the *ehf*-image which is a header-only file type and the variance storage. The way they interact if far from being clear to me.  I believe there should be 2 separated pull-requests.

If *ehf* is different from *edf*, then it should be in a different module. Call it EhfImage and inherit from EdfImage. There is an example with PilatusImage which inherits from TiffImage you can refer to. 

I don't like the tricks of parsing from the beginning or from the end depending on the content type.
Referring to the Zen of Python: 
**Explicit is better than implicit**
I would suggest you to create two lists or dicts, one for signal and one for variance where you indicate what is where. Beside this I would never call a variance ERROR ! That's just misleading for the guy who comes next, regardless who he is.

boesecke:
This pull request can be closed. It is partially replaced by the pull request #318 which prepares the structure of the edfimage reader. More commits for approaching the required ADDITIONAL functionality will follow later.

kif:
OK

PR #315, opened 2019-05-21
Robustness on EDF

vallsv:
This branch was remaining in my workspace.

It was from a previous feed back from P. Boesecke.

- Use `blobsize` instead of `size`
- Size is reserved for the decompressed size
- Add tests on `Sphere2Saxs` data output


kif:
Looks good. Thanks for the PR

PR #317, opened 2019-06-03
Makes filename series split with $ in the name

t20100:
This PR changes the way pixiimage file series filename is split to work even with more than on `$` in the filename.


kif:
Looks good

PR #318, opened 2019-06-07
Reading edf files, without all restrictions of the "Klora" convention.

boesecke:
Reading edf files, without assuming all restrictions of the "Klora" convention. It prepares the possibility of using a general header (see http://wikiserv.esrf.fr/bliss/index.php/ESRF_Data_Format). The following bug has been removed: A header end pattern that is split across read buffers is recognized. A test has been added (EdfBlockBoundaryCases in test_edfimage1.py). Apart from the bug-fix there no additional features have been added. All existing edf-files should be readable as before, probably with updated messages.

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -125,6 +125,97 @@ DEFAULT_VALUES = {}
 # I do not define default values as they will be calculated at write time
 # JK20110415
 
+# PB38k20190607:
+#        The next three functions generalize and replace code in
+#        EdfFrame._extract_meta_data. They could be defined as
+#        internal methods of the EdfFrame class.
+def get_data_rank(header=None,capsHeader=None):
+    '''
+    Get the rank of data array by searching for DIM_nn with the highest index
+    in the header dictionary. Missing DIM_NN are set to 1, except DIM_1
+    which has the default 0.
+
+    :param: dict header
+    :param: dict capsHeader (optional)
+    :return: int rank
+    '''
+    if header is None:
+        header={}
+    if capsHeader is None:
+        capsHeader={}
+        for key in header:
+           capsHeader[key.upper()] = key
+    rank=0
+    if capsHeader is not None:
+        rank=0
+        for key in capsHeader:
+            if key[0:4]=="DIM_":
+                try:
+                    index=int(key[4:])
+                except ValueError:
+                    logger.error("Unable converting index of {} to integer.".format(key))
+                if index>rank:
+                    rank=index
+    return(rank)
+
+def get_data_shape(rank=0,header=None,capsHeader=None):
+    '''
+    Returns a tuple with the number of dimensions up to the given rank.
+    The dimensions DIM_i are read from the header dictionary. If
+    the corresponding value is missing, the following defaults are
+    used: DIM_1=0, all others 1.
+    The shape tuple is filled from the end to the beginning with the values
+    of DIM_i, i.e. shape[0] is equal to value of DIM_rank, shape[rank-i] is
+    equal to the value of DIM_i (e.g. for rank==2, shape[0]==value(DIM_2),
+    shape[1]==value(DIM_1)).
+
+    :param: int rank
+    :param: dict header
+    :param: dict capsHeader (optional)
+    :return: tuple shape
+    '''
+    if rank is None:
+        rank = 0
+    if header is None:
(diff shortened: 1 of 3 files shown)


fabio/edfimage.py
```diff
@@ -255,27 +313,51 @@ def _extract_header_metadata(self, capsHeader=None):
...
-            # Preset with the calculated size, will be updated after decompressing
+            # preset with the calculated size, will be updated later
+            # with a better value when it becomes available, e.g.
+            # after decompression
             self.size = calcsize
 
+        #+++++++++++++++++++++++++++++
+        # PB38k20190607: ATTENTION, weird!:
+        # little_endian^=LowByteFirst, big_endian^=HighByteFirst
+        # Why should _data_swap_needed depend on bpp?
```
kif: Hi Peter,
The code is not the right place for discussing implementation details. Please use the pull-request facility for this in order to keep the code clean an tidy.

About this topic, little- & big-endian differ only if the datatype has a length longer then 1

fabio/edfimage.py
```diff
@@ -125,6 +125,97 @@
 # I do not define default values as they will be calculated at write time
 # JK20110415
 
+# PB38k20190607:
+#        The next three functions generalize and replace code in
+#        EdfFrame._extract_meta_data. They could be defined as
+#        internal methods of the EdfFrame class.
+def get_data_rank(header=None,capsHeader=None):
```
kif: Please go ahead an make them static functions.

fabio/edfimage.py
```diff
@@ -287,13 +369,18 @@ def _extract_header_metadata(self, capsHeader=None):
...
                 self._data_swap_needed = False
 
-    def parseheader(self, block):
+    #PB38k20190607:
+    # This function has been renamed from parseheader to _parseheader,
+    # because it returns the capitalized header and not the
+    # blobsize, which should be taken from _read_header_block or
+    # if not returned there explicitely from capsHeader after
+    # calling _parseheader.
+    def _parseheader(self, block):
```
kif: Those comments have nothing to do in the source code. It should be explained in the commit message.

fabio/edfimage.py
```diff
@@ -625,16 +715,29 @@ def check_header(header=None):
             new[toAscii(key, ";{}")] = toAscii(value, ";{}")
         return new
 
+    # PB38k20190607: _read_header_block reads a single header and
+    # interpretes immediately the EDF_ specification keys.
+    # These values (header_size, binary_size) must be defined
+    # at the beginning of the header and are interpreted and
+    # returned as additional return values
+    # For the moment no change, only EDF_HeaderSize is interpreted.
     @staticmethod
```
kif: Comments are useless, please complete the docsting so that this content is properly indexed and the documentation properly populated.

kif: When replying 4 values I would advice you to use a NamedTupple:
from collection import namedtupple and define a ReturnType
ReturnType = namedtuple("ReturnType", "header_block header_size, binary_size")

fabio/test/codecs/test_edfimage1.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    Checks the shape of the data arrays.
+    Compares the mean value of each data frame array, with the number given in avglist.
+    filename: name of file to test
+    avglist: list of average values for each frame, stops, if not equal
+    keylist: list of keys to read, stops, if not available
+    If avglist or keylist is shorter than frameno, the last value in the list is used.
+    """
+    global ROOT
+    filename = os.path.join(ROOT, filename)
```
kif: Using global variables is BAD. You should rewrite your function using a method and share this ROOT.

fabio/edfimage.py
kif: Done

fabio/test/codecs/test_edfimage1.py
kif: done

kif:
The failure on py3.4 is indeed linked to the release of lxml which was released Jun 10, 2019

kif:
We are still testing python3.4 under windows. python 3.4 makes sense on debian8 but not on windows computer where people are rather likely to use python 3.6 or 3.7. I think it is time to upgrade our appveyor policy (to be consistent with silx).

PR #322, opened 2019-06-19
Fix edfimage example

juliiaa28:
I tried the previous solution (fabio.edfimage()) and it was not working. I have updated it from the documentation found in the fabio.edfimage module.


kif:
Thanks Julia for reviewing our doc.

PR #323, opened 2019-07-02
Do not create a temporary string (it slows down writing on windows)

jonwright:
Presumably a windows thing - it takes 50 ms to run this tostring line for a 2048x2048 image
Issue noticed on python3.7 64 bits and windows 10.


kif:
Hi Jon,
Looks good. Thanks for heavily testing on windows. It is an important platform for our users but less for us developers :(

PR #328, opened 2019-07-26
Recognizing a Global Header Section and using it as default

boesecke:
The Global Header Section (general block) is the one that starts with the keyword EDF_DataFormatVersion, otherwise no Global Header Section is defined for the opened file. Its contents is used as default for all frames. Checks have been added in test_edfimage1.py for verifying the correct reading of these type of files and for verifying that a keyword of the general block is really used as default in data frame headers.
(originally, "VersionNumber" was defined for this purpose. Because it could be located somewhere inside the header it would require reading the full header. This is time consuming and has been abandoned.)

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -126,7 +126,7 @@ DEFAULT_VALUES = {}
 # I do not define default values as they will be calculated at write time
 # JK20110415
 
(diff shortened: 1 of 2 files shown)


kif:
Looks good to me.

PR #329, opened 2019-07-26
first attempt of a soleil reader which can read a stack of images.

picca:
Hello, can you tel me if this is ok for you

thanks

Fred, with this I can read a soleil file a see a stack of image out of the box from fabio_viewer


fabio/openimage.py
```diff
@@ -219,27 +225,33 @@ def _openimage(filename):
...
 
-    klass_name = "".join(filetype) + 'image'
+    objs = []
+    for filetype in filetypes:
+        klass_name = "".join(filetype) + 'image'
+        try:
+            obj = fabioformats.factory(klass_name)
+            objs.append(obj)
+        except (RuntimeError, Exception):
+            pass
```
kif: We should a least log the fact there was an exception ...
Beside this  Runtime derives from Exception ...
```
In [4]: RuntimeError.__mro__
Out[4]: (RuntimeError, StandardError, Exception, BaseException, object)
```
So `except Exception as err` should be enough

kif:
Can you provide us with a test and a reference test-image ?

PR #339, opened 2019-09-05
Make EDF magic it readable

vallsv:
This PR make the EDF magic readable.

As result, we can easily see that there is dup:
- b"\n{\r\nEDF" is already triggered by b"\n{"
- b"{\r\nEDF" is already triggered by b"{"
- b"{\n"  is already triggered by b"{"
- b"\n{\n",'edf') is already triggered by b"\n{"

I let you see what to do with this information.


kif:
Thanks !

PR #341, opened 2019-09-27
Activate PPC64le tests

kif:


kif:
@vallsv 
what do you think of that ? is the compilation too long ?

Issue #316, opened 2019-05-21
Disable numpy allow_pickle


vallsv:
It would be good to disable `allow_pickle=True` in the module numpy image as it is creating vulnerability issues.

https://github.com/numpy/numpy/blob/master/numpy/lib/npyio.py#L316

Test was failing (#313) cause the result of `fabioconvert` was generating numpy images containing pickle data. The PR #314 was fixing the tests. But it is maybe possible to avoid the issue, and to store a single image, or standard types (if supported), instead of pickled objects.

The content of this file should be anyway checked.

picca:
Hello, I just activated the Salsa-CI and the autopkgtests for Debian, and it fails with thie error.

```
======================================================================
ERROR: testFullFormatName (fabio.test.testfabioconvert.TestFabioConvert)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/fabio/test/testfabioconvert.py", line 135, in testFullFormatName
    image = fabio.open("output/03.npy")
  File "/usr/lib/python3/dist-packages/fabio/openimage.py", line 159, in openimage
    obj = obj.read(obj.filename, frame)
  File "/usr/lib/python3/dist-packages/fabio/numpyimage.py", line 161, in read
    self.dataset = numpy.load(infile)
  File "/usr/lib/python3/dist-packages/numpy/lib/npyio.py", line 447, in load
    pickle_kwargs=pickle_kwargs)
  File "/usr/lib/python3/dist-packages/numpy/lib/format.py", line 696, in read_array
    raise ValueError("Object arrays cannot be loaded when "
ValueError: Object arrays cannot be loaded when allow_pickle=False
```

kif:
On Sat, 19 Oct 2019 02:29:46 -0700
picca <notifications@github.com> wrote:

> Hello, I just activated the Salsa-CI and the autopkgtests for Debian, and it fails with thie error.

It has been addressed recenly ...

https://github.com/silx-kit/fabio/commit/ec7a2d124e337cccd2b1f1fb8b2b8f7f08ffcba5

Issue #313, opened 2019-05-21
Tests failing with numpy 1.16.3


t20100:
There is an issue while running `fabio` tests with `numpy` 1.16.3 (tested on both windows and linux with all python):

```
======================================================================
ERROR: testFullFormatName (fabio.test.testfabioconvert.TestFabioConvert)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/python3.4/site-packages/fabio/test/testfabioconvert.py", line 135, in testFullFormatName
    image = fabio.open("output/03.npy")
  File "/python3.4/site-packages/fabio/openimage.py", line 159, in openimage
    obj = obj.read(obj.filename, frame)
  File "/python3.4/site-packages/fabio/numpyimage.py", line 161, in read
    self.dataset = numpy.load(infile)
  File "/python3.4/site-packages/numpy/lib/npyio.py", line 447, in load
    pickle_kwargs=pickle_kwargs)
  File "/python3.4/site-packages/numpy/lib/format.py", line 692, in read_array
    raise ValueError("Object arrays cannot be loaded when "
ValueError: Object arrays cannot be loaded when allow_pickle=False

----------------------------------------------------------------------
```

Same test pass with numpy 1.15.0

t20100:
Reproduced by re-running fabio CI: https://travis-ci.org/silx-kit/fabio/jobs/512908095

t20100:
It was also working with 1.16.2... It seems to be broken only by v1.16.3.

PR #344, opened 2019-10-09
fixed issue #316 and issue #313

fpwg:
The NumpyImage class writes the data in its dataset attribute to the file rather than using its data attribute which is used by the conversion function of FabioImage. The NumpyImage writes the data attribute to the file (which is None in that case) and creates a pickeled file without the actual data in it.

### DIFF
--- a/fabio/numpyimage.py
+++ b/fabio/numpyimage.py
@@ -158,7 +158,7 @@ class NumpyImage(fabioimage.FabioImage):
         self._readheader(infile)
 
         # read the image data
-        self.dataset = numpy.load(infile, allow_pickle=True)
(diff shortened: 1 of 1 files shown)


fpwg:
There could be other cases with similar problems so there should be some new cases that actually check if there is data in the converted files

kif:
Perfect.

vallsv:
Thanks for the PR.

I guess to problem happen when people edit manually `dataset` attribute? In this case it could be convenient to create instead a get/set property for it, and make the attribute private. Maybe this dataset should be use use outside, but i dont have the big picture.

Issue #346, opened 2019-10-10
Add support for CrysAlis Pro Esperanto file format


fpwg:

fpwg:
Closed by #347

PR #347, opened 2019-10-10
add support for esperanto files (issue #346)

fpwg:
adds support for esperanto files. Testing and test data as well as documentation are needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,6 @@ nosetests.xml
 .pydevproject
 .settings
 
-# generated c files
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -37,3 +37,6 @@ nosetests.xml
 .project
 .pydevproject
 .settings
+
+# generated c files
+*.c
```
kif: *.c files should be ignored in fabio/ext but not in fabio/ext/src where pure C files are located.

PR #351, opened 2019-10-29
Work on Esperanto format ...

kif:
Disable this format on Python2


kif:
I leave as it is ... there is a warning about compressed data which may be incorrectly decompressed.

PR #355, opened 2020-01-10
Add debian11/ubuntu20.04 packaging

t20100:
This PR adds packaging for debian11 bullseye (and ubuntu 20.04).

`package/debian11` is made from a copy of `package/debian10` modified to drop python2 (see commit afdf08c).
The `build-deb.sh script is also cleaned up by removing code for debian < 9.


kif:
I wish we keep the "stdeb" route as it is a fail-safe option which can be useful in odd cases.

t20100:
Restored `stdeb` with `--stddeb-py3` and `--stdeb-py2.py3` options.

kif:
Thanks !

PR #357, opened 2020-02-21
Add changelog and prepare a release

t20100:
This PR adds a `CHANGELOG.rst` file and bump the version to `0.10.0` in order to prepare a release of `fabio`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,179 @@
+Change Log
+==========
+
+0.10.0:
+-------
+
+- `edfimage`:
+
+  - Improved robustness (PR #315)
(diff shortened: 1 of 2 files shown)


t20100:
I made the 0.10.0 changlog from the PR descriptions, it needs to be double checked.
For the other versions, I copied what's in the github release descriptions.

kif:
I am surprised that the changelog is not seen as part of doc/sources/ ... that's where it is stored

vallsv:
Yes, here it is https://github.com/silx-kit/fabio/blob/master/doc/source/Changelog.rst

vallsv:
But i agree with thomas, it's better to move it to the root.
Then we should merge this documents.

PR #369, opened 2020-04-06
Drop Python2 for real

### LINKED ISSUE #360: Remove Python2 compatibility code
While this code is left for version 0.10, it should be removed for 0.11, here are a few examples:

* [x] `from __future__ import `
* [x] `from six import`
* [x] Stop using unicode, prefer strings where needed 
* [x] drop argparse from third party as it is available from py2.7+
* [x] drop gzip from third party as it is available from py2.7+

kif:
Close #360


vallsv:
Except my comment, it's fine.

Also `python3-six` could be removed from the debian packages?

vallsv:
I have fixed the conflict

vallsv:
Nice clean up BTW

kif:
No new comments ... then I will merge

PR #371, opened 2020-04-07
mistake tobytes vs tostring

### LINKED ISSUE #370: openimage(FilenameObject) broken in 0.10.0 on github
fabioutils.FilenameObject has a tostring() method but not a tobytes() method. I would be tempted to revert the edit in openimage, but the bytes/string story is not an area I can really help with. If you have unicode filenames then open probably need a string argument??

This broke very recently so I am a bit surprised to find it in the 0.10.0 release file from github.
```
  File "/home/esrf/wright/miniconda3power9/envs/pow37/lib/python3.7/site-packages/fabio/openimage.py", line 151, in openimage
    logger.debug("Attempting to open %s" % (filename.tobytes()))
AttributeError: 'FilenameObject' object has no attribute 'tobytes'
```

kif:
close #370

### DIFF
--- a/fabio/openimage.py
+++ b/fabio/openimage.py
@@ -52,7 +52,6 @@ from .fabioimage import FabioImage
 # Make sure to load all formats
 from . import fabioformats  # noqa
 
-
 MAGIC_NUMBERS = [
     # "\42\5a" : 'bzipped'
     # "\1f\8b" : 'gzipped'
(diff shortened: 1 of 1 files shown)


kif:
@jonwright would this address your issue ?

jonwright:
It looks right to me. I will try to figure out some testcases from the various fable codes.

PR #374, opened 2020-04-08
testcase for v0.10.0 -> v0.10.1 FilenameObjects

jonwright:


jonwright:
So I see the CI is failing but I am not sure why: it seems to be a problem of finding testdata? On my system it fails on a checkout of v0.10.0 and passes on v0.10.1.

vallsv:
Looks like problem with gzip, bzip. @kif any idea? Is it about the remove of the patch of the gzip lib?

vallsv:
Maybe a missing dependency?

jonwright:
I think I am doing a bad tearDown - I will try again...

PR #375, opened 2020-04-08
testcase for FilenameObjects

jonwright:
I will try to push this to v0.10 branch as I assume I pushed to master and that was wrong. My apologies for being rather bad at git.

### DIFF
--- a/fabio/openimage.py
+++ b/fabio/openimage.py
@@ -52,7 +52,6 @@ from .fabioimage import FabioImage
 # Make sure to load all formats
 from . import fabioformats  # noqa
 
-
 MAGIC_NUMBERS = [
     # "\42\5a" : 'bzipped'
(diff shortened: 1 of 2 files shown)


jonwright:
So I was doing bad tearDown (my /tmp folder grows fabio folders, but I guess that is not this issue?).
Unrelated, a git checkout is 123 MB, so some test data are in .git but we are grabbing them over the network and adding/removing in /tmp for each test.

jonwright:
Looking at the CI failure it seems it times out getting a test frame for 32 bit in an unrelated test. The intention is to try to protect the use cases we have over in fable, such as `fabio.open( FilenameObject )` but maybe I messed up somehow?

vallsv:
That's good to me. Be careful i have patched your github branch.

vallsv:
And i rebased it to the master.

PR #376, opened 2020-04-08
Edf badbinary

### LINKED ISSUE #373: Read error on non-ascii characters for edf headers
I found an edf image with some bad bytes (0xcd) padding the header. It was written by an old windows device server so probably was unititialized memory. I cannot them read with fabio (0.9.0 or 0.10.0). Not sure when it stopped working but we could read them in the past. Here is a stripped testcase:
```
>>> import fabio
>>> fabio.edfimage.edfimage( numpy.zeros((20,24), numpy.uint16) ).write("/tmp/test.edf")
>>> f=bytearray(open("/tmp/test.edf","rb").read())
>>> f[512-20]=0xcd                                          # mimic my bad data
>>> open("/tmp/testbad.edf","wb").write(f)
>>> fabio.open("/tmp/testbad.edf")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/fabio/openimage.py", line 159, in openimage
  File "/usr/lib/python3/dist-packages/fabio/edfimage.py", line 788, in read
    raise e
  File "/usr/lib/python3/dist-packages/fabio/edfimage.py", line 778, in read
    self._readheader(infile)
  File "/usr/lib/python3/dist-packages/fabio/edfimage.py", line 722, in _readheader
    block = self._read_header_block(infile, len(self._frames))
  File "/usr/lib/python3/dist-packages/fabio/edfimage.py", line 702, in _read_header_block
    return block[begin_block:end_block].decode("ASCII")
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcd in position 491: ordinal not in range(128)

```
The real frames are in rnice:/data/id11/3dxrd/inhouse/jon/mar10/fe3o4_tomo_a_/

Sadly I am out of my depth with python strings. One option seems to be more forgiving:
`    return block[begin_block:end_block].decode("ASCII" , errors='replace') `
... but then I guess you corrupt strings when people have stored filenames in headers. In that case I guess someone might want bytes without any decoding and go figure it out themselves. So just delimit start and end with "{" and "}" and then split on ";" and "=".

Somewhat related. I still didn't figure out how run individual testcases, only the whole suite and then the testdata disappear. Maybe another issue for that?

jonwright:
These are based off of tag v0.10.1 so I will try to push there. It is supposed to fix #373. Please feel free to disagree with the approach and propose something else.

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -1022,16 +1022,10 @@ class EdfImage(fabioimage.FabioImage):
                 key, val = line.split(b'=', 1)
                 key = key.strip(bytes_whitespace)
                 val = val.strip(bytes_whitespace)
-                # do the decode here for each item:
                 try:
-                    key = key.decode("ASCII")
+                    header[key.decode("ASCII")] = val.decode("ASCII")
                 except:
-                    logging.warn("Non ascii data in your header %s"%(key))
-                try:
-                    val = val.decode("ASCII")
-                except:
-                    logging.warn("Non ascii data in your header %s"%(val))
-                header[key] = val
+                    logger.warning("Non ASCII in header: %s %s", repr(key), repr(val))
 
         # Read EDF_ keys
         # if the header block starts with EDF_DataFormatVersion, it is a general block
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -1007,25 +1007,20 @@ class EdfImage(fabioimage.FabioImage):
         # Go to the start of the binary blob
         infile.seek(offset, os.SEEK_CUR)
 
-        # keep header_block as bytes for issue #373
-        header_block = block[begin_block:end_block]
+        header_block = block[begin_block:end_block].decode("ASCII", errors='ignore')
 
         # create header
         header = OrderedDict()
 
         # Why would someone put null bytes in a header?
-        bytes_whitespace = (string.whitespace + "\x00").encode('ASCII')
+        whitespace = string.whitespace + "\x00"
 
         # Start with the keys of the input header_block
-        for line in header_block.split(b';'):
-            if b'=' in line:
-                key, val = line.split(b'=', 1)
-                key = key.strip(bytes_whitespace)
-                val = val.strip(bytes_whitespace)
-                try:
-                    header[key.decode("ASCII")] = val.decode("ASCII")
-                except:
-                    logger.warning("Non ASCII in header: %s %s", repr(key), repr(val))
+        for line in header_block.split(';'):
+            if '=' in line:
+                key, val = line.split('=', 1)
+                key = key.strip(whitespace)
+                header[key] = val.strip(whitespace)
 
         # Read EDF_ keys
         # if the header block starts with EDF_DataFormatVersion, it is a general block
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -1007,20 +1007,31 @@ class EdfImage(fabioimage.FabioImage):
         # Go to the start of the binary blob
         infile.seek(offset, os.SEEK_CUR)
 
-        header_block = block[begin_block:end_block].decode("ASCII")
+        # keep header_block as bytes for issue #373
+        header_block = block[begin_block:end_block]
 
         # create header
         header = OrderedDict()
 
         # Why would someone put null bytes in a header?
-        whitespace = string.whitespace + "\x00"
+        bytes_whitespace = (string.whitespace + "\x00").encode('ASCII')
 
         # Start with the keys of the input header_block
-        for line in header_block.split(';'):
-            if '=' in line:
-                key, val = line.split('=', 1)
-                key = key.strip(whitespace)
-                header[key] = val.strip(whitespace)
+        for line in header_block.split(b';'):
+            if b'=' in line:
+                key, val = line.split(b'=', 1)
+                key = key.strip(bytes_whitespace)
+                val = val.strip(bytes_whitespace)
+                # do the decode here for each item:
+                try:
+                    key = key.decode("ASCII")
+                except:
+                    logging.warn("Non ascii data in your header %s"%(key))
+                try:
+                    val = val.decode("ASCII")
+                except:
+                    logging.warn("Non ascii data in your header %s"%(val))
+                header[key] = val
 
         # Read EDF_ keys
         # if the header block starts with EDF_DataFormatVersion, it is a general block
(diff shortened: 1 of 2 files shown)


vallsv:
Looks fine, but i don't think we should store bytes in the header.
So at worst i think it is better to skip the key/value pairs. If it is displayed in the log that's maybe enough for people who want to restore it, else surrogate, but as you said, that's python 3 only.

Also please use `logger.warning("fooo %s", value)`
- Use the logger (not logging)
- Fullname `warning`
- No `%` formatting, but a coma.

While we are here, it would be good to also add a warning to the missing `else` if there is no '=' in the line. This data is also skipped.

We can fix it for you if you like.

What do you think @kif ?

jonwright:
Skipping and warning sounds good, I would like to try again with the pull to try to overcome my fear of using git. People will lose their emoji, they should not have put them in edf!

vallsv:
You also can use the github editor, but it's like cheating :-)

jonwright:
I left off the printing for missing "=" items as it can make a lot of noise (many 0xcd bytes in a few hundred files). You might want to change this but I did not find an idiom to only print a warning the first time it happens?

kif:
Is there one rule which says "decode early, encode late". As nobody was able to apply it, this rule has been hard-coded into Python3.
Jon, what you are suggesting is  against this.

If raising an error (Which I believe is correct) does not fit you an you believe the garbage content is just padding, why not using decode("ascii", errors="ignore") ?

jonwright:
Seems like another good option to me. So long as the binary gets through I should be happy. 

I was concerned about pyFAI-average storing unicode filenames in edf headers, but this is not a priority.

jonwright:
Seems there is a utf-7 encoding that someone could use if they really want to put unicode in an edf header. I see how anyone can read it without knowing they did that?

vallsv:
It's pointless, we already can put utf-8 in the header without problems

jonwright:
Right now I just want to read old data which had wrong pad bytes at the end of the header. This part needs a rapid backport onto a release.

Unicode and teaching me python3 is a harder problem even if it can save you all time in the long run: I tried to put utf-8 and latin-1 into fabio headers to make the testcases and did not figure out how to do it? This is why I used bytearray instead. Reading the existing code I understood edfimage.write is forcing 7-bit encodings to be compatible with @boesecke specifications and so if reading does the same it seems OK (e.g. drop the bad characters, warning or not). So far as I can tell, the edfimage constructor silently drops characters while the writer raises errors if you put them into the header later. All that behavior could be cleaned up in the future if someone needs it, but clearly should not be me touching it.

vallsv:
I don't think this fix will anyway be part of a release soon. As you said it was already not working with fabio 0.8. Can't you just use fabio 0.7.

I was not aware of 7-bits limitations. Sounds weird. But all i know is we can't be consistent with ascii and utf7 at the same time without storing the used encoding in the file.

vallsv:
And i think your previous PR was better, ignoring errors is really not a good idea.

jonwright:
The last tag I find working is v0.1.3 which was for python 2.5. Since v0.2.0 it has been broken and nobody noticed until now. The data in question were written by the Quantix device server from @andygotz which ran at ID11 for many years.

My feeling is that ESRF has a duty to provide software to read ESRF data that are in edf format without asking users to fix the files. No idea how best to do that however!

boesecke:
Dear Jon,

could you send such a file, please? It could be possible adding exceptions 
(not nice) to the reader. In any case, using non-ASCII encodings in edf 
headers could break many readers if they determine the length of the 
header in the usual way by searching a specific marker byte that is a 
closing '}' (in ASCII).

Greetings,
Peter

On Thu, 9 Apr 2020, Jon Wright wrote:

> 
> The last tag I find working is v0.1.3 which was for python 2.5. Since v0.2.0 it has been broken and nobody noticed until now. The data in
> question were written by the Quantix device server from @andygotz which ran at ID11 for many years.
> 
> My feeling is that ESRF has a duty to provide software to read ESRF data that are in edf format without asking users to fix the files. No
> idea how best to do that however!
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or
> unsubscribe.[ALY2ZWS2ZHNIDUJWG6CZ4TLRLW333A5CNFSM4MEEADKKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOERZK32Y.gif]
> 
> 
>

================================================================================
Dr. Peter Boesecke
ID01 & ID02

Tel. +33 (0) 4 7688 2400
Fax. +33 (0) 4 7688 2542
Email boesecke@esrf.fr

Postal address
ESRF - The European Synchrotron
CS 40220
38043 GRENOBLE CEDEX 9, France

Physical address
ESRF - The European Synchrotron
71, Avenue des Martyrs
GRENOBLE, France
================================================================================

jonwright:
Hi Peter, 
I put one here:
http://ftp.esrf.fr/tmp/fe3o4_tomo_a__quantix_0227.edf
I put a couple of proposed code changes on github to fix the problem but it seems to be not-so-clear what to do for the best
Thanks!
Jon

jonwright:
Pymca and many other programs also read the file:
```gnuplot> plot "fe3o4_tomo_a__quantix_0227.edf" binary filetype=auto with image ```
```octave> a = edfread('fe3o4_tomo_a__quantix_0227.edf') ; ```
Gui softwares include: ImageJ (edfreader plugin), Fit2d (Klora), Dawn, and XOP/Xplot2d (IDL).

So I think this file was accepted by fabio from 2007 up to the python3 port in early 2015. It would be good to get a fix on a version that runs with both 2 and 3 (e.g. v0.10.x).

jonwright:
I suggest not to merge yet : if we just decode edf headers using an 8 bit table  (latin-1) rather than 7-bit ASCII then I think all can be resolved cleanly.

vallsv:
No that's not a good idea at all.

It's not because there is no exception, that there is no errors.

I will take sometimes today or tomorrow to rework what you have done. I think your previous approach was the good one. I will just clean up it.

BTW, i think 0.10 is already Python 3 only.

jonwright:
Thanks! The latin-1 idea comes from here: http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html

jonwright:
Thanks! The latin-1 idea comes from here: http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html

vallsv:
I was not able to push on this branch, so i created another PR.
Superseded by #381.

Issue #377, opened 2020-04-09
Context error while using BytesIO as input


vallsv:
Feedback from ID10.

They try to read edf from an archive.

Fabio is supposed to work with `ByteIO`, but a context manager looks broken.

```
with ZipFile(my_file,'r') as myzip:
for filename in myzip.namelist():
if filename.find(".edf") != -1:
print(filename)
#in_memory.write(myzip.read(filename))
in_memory = io.BytesIO(myzip.read(filename))
in_memory.seek(0)
print(in_memory.read())
in_memory.seek(0)
f = EdfFile.EdfFile(in_memory)
data = f.GetData()
#data = loadedf(in_memory)
print(data.shape)
plt.imshow(data,vmax=2)
plt.show()
raw_input()
exit()
```

```
Traceback (most recent call last):
File "./read_zip.py", line 71, in <module>
f = fabio.open(in_memory)
File "/users/chushkin/.local/lib/python2.7/site-packages/fabio/openimage.py", line 157, in openimage
obj = _openimage(filename)
File "/users/chushkin/.local/lib/python2.7/site-packages/fabio/openimage.py", line 200, in _openimage
with imo._open(actual_filename) as f:
AttributeError: BytesIO instance has no attribute '__exit__'
```

vallsv:
Closed  by #378

PR #378, opened 2020-04-10
Fix BytesIO

vallsv:
The `_openimage` was never tested.

There is a regression in 0.10, so it was not possible to read `BytesIO` with `fabio.open`.

This is only a fix which could be part to a bug fix.

The `_openimage` should be reworked. This code is reading twice the streams (read the full data for the magic, then read again the full data to parse the stream). This should be avoided, if you deal will standard compressed streams, it's just not efficient, and maybe buggy.

### DIFF
--- a/fabio/openimage.py
+++ b/fabio/openimage.py
@@ -188,10 +188,12 @@ def _openimage(filename):
 
     """
     if hasattr(filename, "seek") and hasattr(filename, "read"):
-        # Looks to be a file containing filenames
-        if not isinstance(filename, BytesIO):
-            filename.seek(0)
-            actual_filename = BytesIO(filename.read())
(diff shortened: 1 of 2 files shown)


vallsv:
Closes #377

kif:
LGTM. Thanks for the input.
We have to ensure fabioutils.BytesIO behaves also as a context manager

vallsv:
I did the test for that, it is working. I suppose it was not working only in Python2 (cause of the inheritance, i guess)

kif:
On Tue, 14 Apr 2020 05:24:40 -0700

> I did the test for that, it is working. I suppose it was not working only in Python2 (cause of the inheritance, i guess)
perfect then.

PR #381, opened 2020-04-11
Robustness on EDF header

### LINKED ISSUE #373: Read error on non-ascii characters for edf headers

vallsv:
This PR is a clean up of #376 from @jonwright 

- warning message in case of non-ascii char in the header (in this case the key-value is drop)
- warning on duplicated key (same behavior as before, the previous key is drop)
- debug message on header line without = (not a key-value)
    - i think it would be definitely better to log it as warning

Closes #373

### DIFF
--- a/fabio/edfimage.py
+++ b/fabio/edfimage.py
@@ -1006,20 +1006,31 @@ class EdfImage(fabioimage.FabioImage):
         # Go to the start of the binary blob
         infile.seek(offset, os.SEEK_CUR)
 
-        header_block = block[begin_block:end_block].decode("ASCII")
+        # keep header_block as bytes for issue #373
+        header_block = block[begin_block:end_block]
 
         # create header
         header = OrderedDict()
 
         # Why would someone put null bytes in a header?
-        whitespace = string.whitespace + "\x00"
+        bytes_whitespace = (string.whitespace + "\x00").encode('ASCII')
 
         # Start with the keys of the input header_block
-        for line in header_block.split(';'):
-            if '=' in line:
-                key, val = line.split('=', 1)
-                key = key.strip(whitespace)
-                header[key] = val.strip(whitespace)
+        for line in header_block.split(b';'):
+            if b'=' in line:
+                key, val = line.split(b'=', 1)
+                key = key.strip(bytes_whitespace)
+                val = val.strip(bytes_whitespace)
+                # do the decode here for each item:
+                try:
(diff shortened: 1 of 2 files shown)


vallsv:
Test are passing, there is just a git problem in one of the Travis CI.

vallsv:
@kif could you take look at that, so i can backport to 0.10 branch?

PR #382, opened 2020-04-14
Declare only python3 comparibility

### LINKED ISSUE #379: version confusion
My CI for ImageD11 is getting the fabio-v0.10.0 release from pip/pypi for python2.7. Since master started getting breaking changes can we add a Requires-Python please?
https://packaging.python.org/guides/dropping-older-python-versions/

kif:
To avoid any confusion, let's state it clearly and prevent Python2 from installing fabio 0.11
close #379 

@jonwright, is this OK for you ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "03/04/2020"
+__date__ = "14/04/2020"
 __status__ = "stable"
 
 import sys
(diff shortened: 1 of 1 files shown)


jonwright:
Looks right to me for fixing the next release.

I am concerned about the status of "pip install " for any package with a fabio dependency right now, it seems to be breaking on python2 for people who need binary wheels (e.g. windows users). Various options:
- withdraw/delete v10 so python2 folks fall back to v9 and release a v11 for python3.5+
- fix v10 on pypi to add wheels (and backports of recent bugfixes) for python2
- leave v10 broken for these people and tell them to upgrade to python3.5+ or install fabio v0.9.0

First two are both fine. I would prefer to avoid the last option as people may want to test new results against (very) old analysis scripts. Your last python25 package was pretty good, it would be nice to do the same for python27. If people are looking at back at data/scripts 10 years from now they probably want option 2.

vallsv:
Sounds reasonable.

kif:
@vallsv are you willing to make a v0.10.2 which contains also  binaries for python 2.7 ?

vallsv:
I don't think i have time for it. I can update the branch, but not to packaging it.

PR #385, opened 2020-05-05
Remove the raise at the end of the iteration

### LINKED ISSUE #384: IOError for EigerImage.getframe() when loading last frame in container
when using fabio to unpack images from the dectris eiger proprietory hdf5 container using the eigerimage.py module, an IOError is raised in line 191. this is because the __iter__ method of fabioimage.py calls the .next() method from eigerimage.py, but when the getframe() method is called on the last frame +1 in this method, the else condition in the getframe() method is invoked and raises the IOError("getframe %s out of range [%s %s[" % (num, 0, self.nframes)).

is there a reason for this else statement to exist at all? how can this be fixed?

vallsv:
Here is a PR to fix #384

Considering this API was not used (cause there was this exception) i would prefer to alias if to `frames` instead of using the iteration through filenames.
```
     def __iter__(self):
        for frame in self.frames():
            yield frame
```
What do you think?

### DIFF
--- a/fabio/fabioimage.py
+++ b/fabio/fabioimage.py
@@ -852,7 +852,7 @@ class FabioImage(_FabioArray):
             try:
                 current_image = current_image.next()
             except IOError:
-                raise StopIteration
+                break
 
 
 fabioimage = FabioImage


kif:
This perfect for python >= 3.5
--> not to be backported 
https://stackoverflow.com/questions/14183803/what-is-the-difference-between-raise-stopiteration-and-a-return-statement-in-gen

vallsv:
For me this exception should only have to me raised by `next()` when you do something like `next(generator)`.

But what about my comment with `frames`?

vallsv:
BTW i discover this syntax `yield from line_pair_iterator`

PR #393, opened 2020-06-15
Provide CBF magic bytes and use them

### LINKED ISSUE #388: CBF write method allows filenames that are not compatible with fabio.open
It seems the CBF format does not have a "magic" byte string at the top of the file to identify the file type. If the name is wrong, it is not recognised (see https://github.com/FABLE-3DXRD/ImageD11/issues/98).

```
>>> im = fabio.open("my_raw_data.cbf")
>>> im.data = processed(im.data)
>>> im.write( "my_fixed_data.edf" )
```
... the resulting file is not in edf format, but remains as a CBF, the type that was opened. I guess there needs to be some kind of check on the filename whenever the filename is in conflict with the file format. Or perhaps automatically add it if there is a required filename extension.

kif:
close #388


vallsv:
Sounds nice.

PR #394, opened 2020-06-15
Lima Image

### LINKED ISSUE #356: limaimage
Expose HDF5 files written by LImA

kif:
Provide initial support for Lima generated HDF5 files
close #356


vallsv:
Is that ready to merge?

How do you expect to use it? I guess you will not use `fabio.open` but use explicitly the codec class right?

Maybe it could be nice to think about a kind of dispatcher for this kind of file format shared by many codecs (that's the same problem with TIFF images), a single shared step to open the file, and another specific step per codec to check the file content. Just an idea.

kif:
The distinction between an HDF5 generated by an Eiger detector and a detector controlled by LIMA (which could be an Eiger) is described in
https://github.com/silx-kit/fabio/pull/394/commits/506018ca3721ff79ecf483e1dfaa4d726a1dfade
Apparently it works.

vallsv:
Ah yes i see, there is plugin specific code in the core.
Anyway, can it be merged?

kif:
I think so. Scientists are getting nervous with their new detectors arriving.
This will allow "on the field" testing.

vallsv:
I think it would be good to have the same thing for the eiger. If possible.
Any idea if there is a specific field too?

kif:
On Wed, 17 Jun 2020 00:28:06 -0700

> I think it would be good to have the same thing for the eiger. If possible.
> Any idea if there is a specific field too?

I don't think so :(

PR #399, opened 2020-06-28
Implement and test hdf5format 'next' function

### LINKED ISSUE #397: Getting the next hdf5 frame ?
Trying out Hdf5Image.next() it seems it returns a FabioFrame instead of a FabioImage. This does not have a "next" method for getting the N+1 and so on. Various options would "fix" it for me :
- change the "next" method within Hdf5Image to return something inheriting from FabioImage (as advertised)
- add a "next" method on Hdf5Frame
- have Hdf5Frame inherit from Hdf5Image (smallest change)

"next" seems to be well defined for slices into a 3D array. In the longer term I could try to abandon this "next" functionality, I am not sure if anyone else is using it anyway. Do you have any preference if I try to make a patch?

vallsv:
This PR implements the `next` function to HDF5 frames.

Limitation: As the link between frame to container is a weakref, the reference to the fabioimage have to be referenced by the user.

Closes #397

### DIFF
--- a/fabio/fabioimage.py
+++ b/fabio/fabioimage.py
@@ -380,6 +380,14 @@ class FabioFrame(_FabioArray):
             return self._dtype
         return self.data.dtype
 
+    def next(self):
+        """Returns the next frame from it's file container.
+
+        :rtype: FabioFrame
+        """
+        container = self.file_container
+        return container.get_frame(self.file_index + 1)
+
 
 class FabioImage(_FabioArray):
     """A common object for images in fable
--- a/fabio/hdf5image.py
+++ b/fabio/hdf5image.py
@@ -147,8 +147,8 @@ class Hdf5Image(fabioimage.FabioImage):
         Returns a frame as a new FabioImage object
         :param num: frame number
         """
-        if num < 0 or num > self.nframes:
-            raise RuntimeError("Requested frame number %i is out of range [0, %i[ " % (num, self.nframes))
+        if num < 0 or num >= self.nframes:
+            raise IndexError("Requested frame number %i is out of range [0, %i[ " % (num, self.nframes))
         # Do a deep copy of the header to make a new one
         frame = Hdf5Frame(self, num)
         return frame
--- a/fabio/hdf5image.py
+++ b/fabio/hdf5image.py
@@ -142,6 +142,9 @@ class Hdf5Image(fabioimage.FabioImage):
             raise RuntimeError(err)
         return self
 
+    def _get_frame(self, num):
+        return self.getframe(num)
+
     def getframe(self, num):
         """
         Returns a frame as a new FabioImage object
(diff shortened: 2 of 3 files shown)


jonwright:
This works for me - it would just help to get an "import hdf5plugin" at the top of the file to read compressed data.
Thanks!

vallsv:
I see, but `hdf5plugin` is the responsibility of the application. I don't think it's a good idea to use it somewhere here in fabio.

kif:
Next, but no previous ... I am sure Jon will request a previous...

vallsv:
I would prefer to do it only if somebody request it.

jonwright:
For my peaksearch then "next" is enough (via fabio.file_series.new_file_series). Fabians previous button will not get fixed here, it would use the frame numbers anyway.

For the hdf5plugin, maybe it is a separate issue? (lima/eigerimage also). We only appear to need the import if:
* the format is hdf
* the dataset comes with that kind of compression
* HDF5_PLUGIN_PATH was not carved on the runes
* the needed compression is not already built in with h5py

Anyway, this little script is not giving a useful error message:
```
import fabio, sys
img = fabio.open(sys.argv[1])
print( "nframes", img.nframes, "shape", img.data.shape, "first pixel", img.data[0,0])
```
Output:
```
(fable) Y:\jon\hdftest\lens_difftomo>python test.py scan0020.hdf::/entry/measurement/data
  File "c:\users\wright\downloads\test.py", line 2, in <module>
    img = fabio.open(sys.argv[1])
  File "c:\miniconda3\envs\fable\lib\site-packages\fabio\openimage.py", line 173, in openimage
  File "c:\miniconda3\envs\fable\lib\site-packages\fabio\hdf5image.py", line 136, in read
    self.data = self.dataset[self.currentframe, :, :]
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "c:\miniconda3\envs\fable\lib\site-packages\h5py\_hl\dataset.py", line 573, in __getitem__
    self.id.read(mspace, fspace, arr, mtype, dxpl=self._dxpl)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5d.pyx", line 182, in h5py.h5d.DatasetID.read
  File "h5py\_proxy.pyx", line 130, in h5py._proxy.dset_rw
  File "h5py\_proxy.pyx", line 84, in h5py._proxy.H5PY_H5Dread
OSError: Can't read data (can't open directory)
```
Perhaps something like this ?
```
    try:
        self.data = self.dataset[self.currentframe, :, :]
    except:
        for fid in self.dataset._filters.keys():
             if not h5py.h5z.filter_avail( int( fid ) ):
                  logger.error("dataset compressed with filter %s that is not available (hint: try to import hdf5plugins)"%( fid ) )
```

```

kif:
If previous is not absolutely needed, then it is OK for me.

PR #403, opened 2020-09-14
Parser for Pilatus specific headers

kif:
Implement a converter to CBF from out Eiger files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ addons:
 # Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
 matrix:
   include:
-   - python: 3.5
-     os: linux
    - python: 3.6
      os: linux
    - python: 3.7
--- a/ci/appveyor.yml
+++ b/ci/appveyor.yml
@@ -25,15 +25,6 @@ environment:
         VENV_TEST_DIR: "venv_test"
 
     matrix:
-        # Python 3.5
-        - PYTHON_DIR: "C:\\Python35"
-          PYTHON_VERSION: "3.5.10"
-          PYTHON_ARCH: "32"
-
-        - PYTHON_DIR: "C:\\Python35-x64"
-          PYTHON_VERSION: "3.5.0"
-          PYTHON_ARCH: "64"
-
         # Python 3.6
         - PYTHON_DIR: "C:\\Python36-x64"
           PYTHON_ARCH: "64"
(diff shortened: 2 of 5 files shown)


kif:
Hi Jon,
as this is for ID11 ... you've ben volontered to review :)
For now the work has only started, so nothing to be done before the label "work in progress" is replaced with the "ready to merge". But we can discuss on the implementation.

jonwright:
I am confused: don't we need to use a real dataset and import that data into crysalis to make any progress here ? If they are guessing the module geometry from the detector name it could cause a few problems...

kif:
I tested the import of CBF files generated from Eiger2 data and it works (-> peak picking + indexing).

There are no correction for module geometry yet.

kif:
Jon, As soon  as you have an actual dataset, I would be interested in testing the conversion (down to CrysalysPro)
Cheers,

jonwright:
Sure - do you have the sucrose dataset from the Frelon that Carlotta 
converted already? The rest of the bliss file structure is likely to be 
similar. Our plan is to run the same crystal when we get the detector 
installed. Hopefully soon.
Thanks!

jonwright:
Hi Jerome,
we can use the sucrose dataset in:
/data/id11/nanoscope/blc12327/id11/sucrose/sucrose_interlaced_forward/sucrose_interlaced_forward.h5
to use it to benchmark your conversion process for the frelon data.

The dataset has been collected using an interlaced forward scan, so it 
means that images need to be sorted in the right order.
Step size is 0.25 (1440 images over 360 degrees).
They need to be corrected for the flat field:
/data/id11/3dxrd/inhouse/Frelon21/F21_flat_Oct2016.edf
the spacial distortion:
/data/id11/3dxrd/inhouse/Frelon21/frelon21_mar16.spline
and the dark:
/data/id11/nanoscope/blc12381/id11/sample/sample_dark/sample_dark.h5

Once you have the cbf files I can try to run them in crysalis and see if 
I get the same results for the data integration.
Let me know if something is not clear,
Best,
Carlotta

kif:
Hi Jon & Carlotta,
Thanks for the dataset, I'll use it as a test case. For this, there are 3 main things to do:
* dark, flat and distortion correction. This is the business of pyFAI, not FabIO, hence it won't be addressed in this PR.
* Export CBF files with the proper Pilatus headers so that CrysalisPro is able to understand them
* Understand the interlacing scheme and possibly re-order frames if CrysalisPro does not accept frames interlaced frames.
This PR addresses only the second point.

kif:
TODO:
 * [ ] Append an option to input the wavelength
 * [ ] Append an option to flip up-down, right-left, transpose the image

kif:
TODO list:
 * [ ] Append an option to input the wavelength
 * [ ] Append an option to flip up-down, right-left, transpose the image
 * [ ] Drop Python 3.5 (uses unsupported features)

kif:
Merge now, will work further on crysalis support when Rigaku answers to our questions

PR #411, opened 2020-11-03
Esperanto format reading and writing

kif:
The constrain about size in the range 256-4096 is not
yet enforced ...
Eiger 16Mpix detector is larger than the 4096² shall we crope?

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -37,7 +37,7 @@ deb_name=$(echo "$source_project" | tr '[:upper:]' '[:lower:]')
 
 # target system
 if [ -f /etc/debian_version ]
-then 
+then
     debian_version=$(cat /etc/debian_version | cut -d. -f1 | grep -o '[0-9]*')
     if [ -z $debian_version ]
     then
@@ -164,13 +164,13 @@ build_deb() {
     then
       cp -f dist/${project}-testimages.tar.gz ${build_directory}
     fi
-    
+
     cd ${build_directory}
     tar -xzf ${tarname}
-    
+
     directory=${project}-${strictversion}
     newname=${deb_name}_${debianversion}.orig.tar.gz
-    
+
     #echo tarname $tarname newname $newname
     if [ $tarname != $newname ]
     then
@@ -180,7 +180,7 @@ build_deb() {
       fi
         ln -s ${tarname} ${newname}
     fi
-    
+
     if [ -f ${project}-testimages.tar.gz ]
     then
       if [ ! -h  ${deb_name}_${debianversion}.orig-testimages.tar.gz ]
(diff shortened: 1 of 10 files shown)


jonwright:
Looks right to me. There is some story about the silicon detector thickness (after pixel size) being optional and only present for pixel detectors but I have the feeling this is not handled well in crysalis anyway. The CI problem seems to be unrelated?

jonwright:
If an image is too big - you could rebin as well as cropping - for small molecules it might be a more useful default ?

kif:
The silicon sensor does not land into the `esperanto` file but in the `par` file (or another side-car file managed by crysalis) if I remember properly.
The CI issue is related to h5py 3.0 which apparently was not found. Thomas Kluyver released it a couple of days ago.

kif:
Shall we extend to 256² and crop to 4096² ?

jonwright:
Padding to 256x256 should not cause a problem for anyone. Reducing to 4096 you might upset someone by clipping the edges of their data - how much larger is it? Raising an error for something too large and suggesting a 2x2 rebin seems safer.

They say the sensor thickness belongs with pixelsize but I don't think the current release is reading this properly:
```
PIXELSIZE drealpixelsizex drealpixelsizey [dsithicknessmmforpixeldetector]
drealpixelsizex (M) - pixel size along x in mm
drealpixelsizey (M) - pixel size along y in mm
Note that: drealpixelsizex = drealpixelsizey is a requirement
[dsithicknessmmforpixeldetector] - thickness of Si for pixel detectors; the presence of this number signals a pixel detector. Pixel detector gap zones are marked with -1
```

kif:
At this stage we should be able to read image in `esperanto` format. The decompressor is written in Python so perfomances are not there yet. The next step is to be able to write them in order to have a reference codec before being able to export a full dataset to Crysalis

kif:
@fpwg I finally got esperanto working ... it was just the "endianness" which was assumed in the wrong way. You did a fairly good job at that time. Thanks again.

kif:
At this stage, both compressed and uncompressed data  can be read by FabIO (properly) but only uncompressed data can be imported properly in Crysalis. The compressed data cannot (yet?) be properly imported back into crysalis: headers are apparently correct but not the pixel values. Apparently our understanding of the compression algorithm is still partial.

jonwright:
Did you test against cryio ? With Vadim's implementation we think we can 
import compressed ...

kif:
I tested against Crysalis, that's what you want to use, no ?

* The headers are OK-ish (at least it does not prevent the data from being imported
* The binary data block (intensities) is byte for byte the same as the one generated by the CBF importer from Crysalis (`dc`).
* But the difference is at the very end of the file. Apparently there is a binary trailer I did not manage yet to generate yet.

I suspect this footer contains the position of the start of the lines (or their length or something similar). Maybe there is a also a checksum. The trailer contains only redundant information as the code we already have is able to read and write without loss of information.

jonwright:
> I suspect this footer contains the position of the start of the lines
> (or their length or something similar).

Indeed, it seems to be for parallel readers. You can get the source code 
for Vadim's (in C) on pypi as cryio to see how it is done:

`blob = bytearray(self.makeheader(i) + 
cryio._cryio._esp_encode(self.padded))`

kif:
The code we had was taking the position `zero` at the wrong place. Now it is fixed. Files are properly read by Crysalis.

kif:
So the export works and the mask is automatically generated and taken into account.

Beam center is automatically propagated.

The export is still penalized by the compression speed which represents 80% of the execution time.

jonwright:
Cool - thanks!

PR #420, opened 2020-11-16
This is prliminary support for sparse frame reader

kif:
Current performances with numpy is 6fps.


kif:
Implements noise based on a basic RNG

PR #432, opened 2020-12-27
address newer version of Glymur

### LINKED ISSUE #431: JP2k support Glymur 0.7 & 0.8 but now newer
Current version is 0.9.2

kif:
close #431

### DIFF
--- a/fabio/jpeg2kimage.py
+++ b/fabio/jpeg2kimage.py
@@ -97,11 +97,9 @@ class Jpeg2KImage(FabioImage):
         version = tuple(int(i) for i in glymur.__version__.split(".")[:2])
         if version == (0, 7):
             image = glymur.Jp2k(filename=filename)
-        elif version == (0,8):
+        elif version >= (0,8):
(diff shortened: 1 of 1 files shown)


vallsv:
This check was designed to be sure the code was safe against glymur cause we use private attribute.

The way you check the version now do nnot handle that.

I suggest to display at least a warning if glymur is bigger than what we expect.

By the way, it sounds like the project is dead anyway.

kif:
> This check was designed to be sure the code was safe against glymur cause we use private attribute.

Well it handles v0.7.x  and 0.8.x and works with 0.9.x. (0.9.1, 0.9.2, 0.9.3)
even if not perfect, it is already better than previouly. 

> I suggest to display at least a warning if glymur is bigger than what we expect.

That's possible ... I just packaged all wheels for 0.11
 
> By the way, it sounds like the project is dead anyway.

Glymur was updated last week (01 Jan 2021!) with version 0.9.3

PR #437, opened 2021-02-12
define a flat at the class level to tell if images should have noise

### LINKED ISSUE #436: Sparse data not noisy by default ...
The first image is always generated without noise which is misleading.

kif:
This applies also for the first image!
close #436


kif:
close #436

PR #438, opened 2021-03-02
cleanup leftover print() commands from cbfimage.py

bjoe2k4:
I believe these lines have no purpose anymore and just clutter the output.


vallsv:
Looks good to me.
Thanks a lot for the fix.

The CI looks to fail for anther reason, so i merge it anyway.

PR #442, opened 2021-03-12
close #441

### LINKED ISSUE #441: Issue when iterating an HDF5 dataset ...
Simple demo:
```
import fabio
for f in fabio.open("example.hdf5::entry/data"):
   print(f)
```
breaks at the end of the iteration

kif:
and prevent regression with a test !


vallsv:
You should not print in the loop.
It's 50 lines of noise in case you have to enable verbosity of tests.
Anyway i merge it.

Issue #443, opened 2021-04-09
Export sparse dataset (from Jungfrau detector) to Eiger-like files


kif:
Multiple options are possible:
* [ ] Lima-like HDF5 files
* [ ] Eiger-like HDF5 files
* [ ] NXMx nexus format 
      - preferred by beam-line staff
      - cannot be read back by FabIO !
To be charged to ID29, serial crystallography project.
@vasole any input on this ?

PR #448, opened 2021-04-28
Densification of sparse datasets

kif:
Densification -> LimaImage
related to  #443


vallsv:
Anyway, i think you should use `fabio-densify` as entry point. THis entry points are shared with the while installed libs. `densify` sounds to me too generic and easy to overwrite by accident.

kif:
I can check if it can be merged to the convert part ... reasonable idea.

kif:
I changed the name of the CLI tool to `densify-Bragg` to mirror the `sparsify-Bragg` implemented in pyFAI.

jonwright:
Obviously, I have no objection to any of this. I just didn't understand what it is meant for.

PR #456, opened 2021-07-26
Change README section order and reformulate some paragraphs

### LINKED ISSUE #455: Outdated and obscure README
Having tried to use fabio again after a few months, I found the README was not very helpful:
- It contains outdated information (e.g. the code snippet https://github.com/silx-kit/fabio#transparent-handling-of-compressed-files does not work for fabio 0.11.0)
- I also got the feeling that "Using fabio in your own python programs" was supposed to be a header ? :shrug: 
- The link to the "real" documentation is not visible. It could also be a direct link to the latest documentation.
- The section https://github.com/silx-kit/fabio#design-specifications is not very readable (could use some backtick formatting)
- Scrolling all the way down is not worth the effort as the Installation and Changelog sections only give links (which cannot even be followed)

If a maintainer agrees on these statements, I will be happy to submit a PR

loichuder:
Fix #455 

- Changed order of sections to put forward **Installation**, **Usage** and **Documentation**
- Updated **Usage** examples
- Added a table of contents
- Reformulated **Transparent handling of compressed files** section (was really hard to understand). Please review to see if the sense is still correct.
- **Installation** and **Changelog** links point now to the online doc

I left **Design Specifications** untouched but in my opinion, it should be in the online doc and not in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,21 +14,64 @@ Main websites:
 FabIO is an I/O library for images produced by 2D X-ray detectors and written in Python.
 FabIO support images detectors from a dozen of companies (including Mar, Dectris, ADSC, Hamamatsu, Oxford, ...),
 for a total of 30 different file formats (like CBF, EDF, TIFF, ...) and offers an unified interface to their
-headers (as a python dictionary) and datasets (as a numpy ndarray of integers or floats)
+headers (as a Python dictionary) and datasets (as a numpy ndarray of integers or floats)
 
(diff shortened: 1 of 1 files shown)


vallsv:
I merge it as it is, it's much better.
I also agree with you for the "Design Specifications", but probably this just have to be removed, cause it's part of the API documentation anyway. Better to provide example than implementation details.

PR #465, opened 2021-12-18
FileI0.read and FileIO.write have a limit in size for a single operation

### LINKED ISSUE #464: FileIO size limit
`FileIO.read` and `FileIO.write` may read/write less bytes than requested, in which case you need multiple calls.

For example:
```python
import os
from io import FileIO

MAX = 2 ** 31 - 4096

filename = "test.dat"
nbytes = MAX + 1
print("Data size bytes:", nbytes)

data = b"0" * nbytes
for openctx in (open, FileIO):
    with openctx(filename, "wb") as f:
        nwritten = 0
        n = 0
        while nwritten < nbytes:
            nwritten += f.write(data[nwritten:])
            n += 1
    print(f"\nFile size bytes: {os.stat(filename).st_size} (written in {n} times)")

    with openctx(filename, "rb") as f:
        nbytesread = 0
        nmissing = nbytes
        n = 0
        while nmissing:
            nbytesread += len(f.read(nmissing))
            nmissing = nbytes - nbytesread
            n += 1
        print(f"Read size bytes: {nbytesread} (read in {n} times)")
```

In other words, Fabio cannot handle images larger that 2GB.

woutdenolf:
Closes #464

### DIFF
--- a/fabio/fabioutils.py
+++ b/fabio/fabioutils.py
@@ -487,6 +487,32 @@ class File(FileIO):
         """
         return FileIO.close(self)
 
+    def write(self, readablebuffer):
+        nbytes = len(readablebuffer)
+        nwritten = super().write(readablebuffer)
+        while nwritten < nbytes:
+            # One write operation is limited to e.g. 2 GB
+            nextra = super().write(readablebuffer[nwritten:])
+            if nextra:
+                nwritten += nextra
+            else:
+                break
+        return nwritten
+
+    def read(self, size = -1):
+        buffer = super().read(size)
+        size -= len(buffer)
+        while size > 0:
+            # One read operation is limited to e.g. 2 GB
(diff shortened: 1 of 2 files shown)


woutdenolf:
Not sure why `FileIO` is used as a base class instead of `TextIOWrapper` (which is what python's `open` returns).

woutdenolf:
I can add a test but this would need 2GB of memory, which might be too much for CI. I could use `psutil` to check the memory before running the test. I tested manually.

kif:
Hi Wout,
Thanks for the PR. It would be interesting to have such a test, even if it is skipped for CI.

woutdenolf:
In analogy with other tests I skip these memory-heavy tests based on the log level.

kif:
Great !

PR #478, opened 2022-06-10
Meson

kif:


fabio/ext/meson.build
```diff
@@ -0,0 +1,49 @@
+py = import('python').find_installation()
+dep_py = py.dependency()
```
eli-schwartz, line 2: You already have this defined from the top-level meson.build, so you do not need to re-import it.

meson.build
```diff
@@ -0,0 +1,28 @@
...
+endif
+
+# https://mesonbuild.com/Python-module.html
+py_mod = import('python')
+py = py_mod.find_installation()
+py_dep = py.dependency()
+
+py.install_sources(
+  ['version.py'],
+  #rename: ['_version.py'],
```
eli-schwartz, line 23: You should be able to do this with 

```meson
version_py = configure_file(input: 'version.py', output: '_version.py', copy: true)
py.install_sources(
    version_py,
    pure: false,
    subdir: 'fabio',
)
```

However, you could also move this to fabio/meson.build, use `input: '../version.py'`, and simply add `version_py` to `python_sources`.

eli-schwartz (changes requested):

fabio/ext/meson.build
kif: Thanks, I a complete newbe with meson

kif:
Mind to use a patched version of meson-python:
https://github.com/FFY00/meson-python/issues/80

Beside this, the structure is there. The complete code is not ported to meson, but should be trivial.

Remains to be done: update all utilities around packaging like
* [ ] bootstrap
* [ ] run-tests
* [ ] debian packaging
* [ ] documentation

PR #482, opened 2022-06-27
Draft: Move read header block to public

payno:


payno:
no needed. In fact you can retrieve 'frame 'raw data' position' from 'EdfFrame' `start` and `blobsize`.

PR #493, opened 2022-09-26
Add support for "Big tiff"

pierrepaleo:
Add magic number to support [big tiff](https://www.awaresystems.be/imaging/tiff/bigtiff.html). Normally only the version number has to be changed (42 to 43).

This is far from being perfect, not sure how it should be done. With this change, fabio is able to fall-back on `PIL` and read the data (and so `silx view` works on these kind of files).

### DIFF
--- a/fabio/openimage.py
+++ b/fabio/openimage.py
@@ -57,12 +57,14 @@ MAGIC_NUMBERS = [
     (b"FORMAT :100", 'bruker100'),
     (b"FORMAT :        86", 'bruker'),
     (b"\x4d\x4d\x00\x2a", 'tif'),
+    (b"\x4d\x4d\x2b\x00", 'tif'), # bigtiff, big endian
     # The marCCD and Pilatus formats are both standard tif with a header
     # hopefully these byte patterns are unique for the formats
     # If not the image will be read, but the is missing
     (b"\x49\x49\x2a\x00\x08\x00", 'marccd/tif'),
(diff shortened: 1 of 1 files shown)


vallsv:
Could you also provide a very small file (like 4 pixels), so what we could test it?

pierrepaleo:
Sure, will do. You are right, let's keep the `.git` lightweight !

vallsv:
There is no need, you can upload the image inside http://www.silx.org/pub/fabio/testimages/

pierrepaleo:
Not sure PIL will provide what I'm looking for in the end. It seems PIL returns just the first frame, even for a 3D stack:

```python
from tifffile import TiffWriter, TiffFile
from scipy.misc import ascent
from PIL import Image

fname = "/tmp/img.tif"

img = ascent().astype("f").ravel()[:5*6*7].reshape((5,7,6))
tw = TiffWriter(fname, bigtiff=True)
tw.write(img)

# with PIL
img_pil = Image.open(fname)
print(img_pil.n_frames) # 1
print(img_pil.size) # (6, 7)

# with tifffile
img_tf = TiffFile(fname)
print(len(img_tf.series)) # 1
print(img_tf.series[0].get_shape()) # (5, 6, 7)
```

kif:
thus the test was useful!

vallsv:
To me it's kind of special cause a TIFF is more than multi layer than multi frames.

So it's kind of interpretation.

But while it is tested, it's fine.

PR #499, opened 2022-11-11
12 bit data is no longer assumed 

osholm:
The assumption that data from HiPic is 12 bit data so if values are able 4095 it is a BG corrected image and these values should be negative, has been removed.

### DIFF
--- a/fabio/HiPiCimage.py
+++ b/fabio/HiPiCimage.py
@@ -58,7 +58,7 @@ class HipicImage(FabioImage):
         Read in a header from an already open file
 
         """
-        Image_tag = infile.read(2)
+        Image_tag = infile.read(2).decode()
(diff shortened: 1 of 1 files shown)


osholm:
Hi Jerome,

Sorry, if you have already looked at this. But I am not that comfortable with Github yet. I though I made a pull request to merge some changes in HiPicimage into to the main branch but it appears to me that I might have close the request myself. Therefore, I try to reissue it here.

Best,
Henning

kif:
I never used those files ... so if all tests pass, I am ok with your changes.

osholm:
As far as I can see they pass the test. I do have failures but these, as far as I can see are all related to me missing PIL/jpeg2000.

osholm:
I can see I am still learning - it did pass all tests on Github apparently. :-)

Issue #502, opened 2022-11-29
TifImage unable to read certain 16-bits images


kif:
[frame.zip](https://github.com/silx-kit/fabio/files/10110740/frame.zip)

vasole:
Photometric interpretation is not correct. One can force reading ignoring it.

PR #503, opened 2022-11-29
[IO] Ignore photometric interpretation if not supported.

vasole:
To try to workaround #502


vasole:
TiffIO reads files when the photometric interpretation is missing.

I guess to ignore wrong information is acceptable, but I let you decide.

kif:
Thanks Armado,
I'll make a non-regression test with this image to validate it works ...

PR #516, opened 2022-12-08
Doc

### LINKED ISSUE #500: Sparse files written by LIMA2 are not identified.
The "creator" (attribute of the root) is `b'LIMA2'` and not `'LIMA2'`, this is one issue, the second is that there is no way to distinguish a file with a dense stack from a sparse stack ...

kif:
close #500


osholm (commented):
Hi Jerome, 

I have looked at the changes and the example is working as explained in the text. Sorry, I was so slow. 

Henning

Issue #508, opened 2022-12-02
move code to `src`


kif:
* Prevents the code from being imported at the top level
* Simplifies the versioning (?)

kif:
Needs to change all meson.build -> only after `setup.py` gets removed !

kif:
done

vallsv:
:+1:

Issue #506, opened 2022-12-02
Move debian packaging to meson


kif:

kif:
Debian packaging can be done later on ...

kif:
since setup.py is still present

kif:
Done

PR #532, opened 2023-06-21
Remove deprecated setup.py files

kif:
- Move the source of fabio to `src`
- adapt debian packaging (only for debian12)
- adapt documentation


kif:
close #508 
close #506

PR #550, opened 2023-12-11
Update ExternalResources.py

### LINKED ISSUE #549: Several regression in Python3.12 ... related to unclosed gzip files
Among other ...
https://bugs.debian.org/1056471

kif:
close #549


kif:
This causes test failure on python 3.12 when not test-data are available locally

EdgarGF93 (approved):

Issue #553, opened 2024-03-04
Support for newer GE files?


psavery:
Hi there,

First of all, thank you for providing and supporting this great project! We use `fabio` to support many detector file formats in [hexrd](https://github.com/hexrd/hexrd).

For GE files, our users have typically used ge2 files, which seem to work fine with fabio.

Recently, however, some users have tried using ge3 and ge5 files. Correct me if I'm wrong, but ge3 files might already be supported by fabio (the examples I see have a blank header, so I can't tell what the difference might be with ge2). But ge5 appear to have a different header format. Looks like it contains key-value pairs:

```
{EDF_DataBlockID = 1.Image.Psd ; ^M
EDF_HeaderSize =       8192 ; ^M
HeaderID = EH:000001:000000:000000 ; ^M
ByteOrder = LowByteFirst ; ^M
Num_Images = 1441 ; ^M
DataType = UnsignedShort ; ^M
Image = 1 ; ^M
Dim_1 = 2048 ; ^M
Dim_2 = 2048 ; ^M
Size = 8388608 ; ^M
}
```

Are there any existing plans to support ge5 files? We may contribute support for them to fabio, if not.

Thank you!
Patrick

jonwright:
Hello,
What is the error you get using `fabio.open("some_example_file")`? It looks like it should be recognised and read as EDF. I thought the blank headers should also get read as GE [here](https://github.com/silx-kit/fabio/blob/c1eff28cd17f9bacf035fb4b163becc7dba9f971/src/fabio/openimage.py#L84)
Best,
Jon

psavery:
Hi @jonwright! Thanks, that is helpful! Yes, I see now that it is actually reading the ge5 files as EDF.

The actual issue we encountered is that only 1 frame appears to be read in. For my example, though, there are 1441 frames (see `Num_Images` above), but `img.nframes` only reports 1 frame. Do you know why this might be?

I would share my example file with you, but I would need to get permission first.

psavery:
I looked through the fabio code and stepped through it a bit. It looks like fabio is expecting a header for each frame, but this file only contains one header at the beginning of the file. fabio raises this exception when reading the second frame:

```python
fabio.edfimage.MalformedHeaderError: Header frame 1 does not contain '{'
```

Indeed, looking at the raw file, there is only a header at the beginning of the whole file.

Is this file invalid? Or does fabio need to be updated to read this file correctly?

FYI: this file was generated at APS.

[...]

PR #554, opened 2024-03-14
Add support for GE files with an EDF header

psavery:
This file type started coming from APS in ~2022 with a `.edf.ge5` extension.

In fabio, these were being interpreted as EDF files because of their EDF-like header. With the changes in this PR, the file is interpreted as a GE file instead and the frames are read in correctly.

The added logic is as follows:

1. If the file is identified as EDF, we check the file extension, and if the file extension is `ge` plus a number, it is interpreted as GE instead.
2. We check for EDF header formats in the GE reader, and if we have a match, we read in the header using EDF style and grab the number of frames from there.

I have an ~80 MB dark example... let me see if we can make that public for testing.

Fixes: #553

### DIFF
--- a/src/fabio/GEimage.py
+++ b/src/fabio/GEimage.py
@@ -41,8 +41,12 @@ __licence__ = "MIT"
 import io
 import numpy
 import struct
+from .edfimage import EdfImage
 from .fabioimage import FabioImage
 from .fabioutils import next_filename, previous_filename
+from .openimage import MAGIC_NUMBERS
+
+EDF_MAGIC_NUMBERS = [(x, y) for x, y in MAGIC_NUMBERS if y == 'edf']
 
 GE_HEADER_INFO = [
     # Name, length in bytes, format for struct (None means string)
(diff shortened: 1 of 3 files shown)


psavery:
I can share the dark file, which is ~80 MB! Is that too large for public testing, or is it okay?

jonwright:
Hi Patrick - I would suggest to gzip the file to make it smaller for testing. Normally fabio reads that transparently, but it breaks the format guessing from file extensions....

kif:
Hi, in Fabio, files are distributes as `bz2` format and then decompressed and recompressed in gzip. This saves a lot of disk space. Don't worry too much about the size of that file.

jonwright (commented):
For now it just misses the testcase ... ?

psavery:
I sent a test case by email. We can make that public.

psavery:
I added the file to the GE tests! Let me know if anything further is needed for merging this PR.

jonwright:
So I guess the question is whether anyone would want to support reading dark_before_000428.edf.ge5.bz2, I will ping the private email thread

psavery:
We can potentially modify the regex to allow `.ge5.*` as well.

Also, in this PR, I am assuming that `Num_Images` will be in the EDF header, since it was there in our example files. I wonder if it's safer to just figure out the number of frames by the file size instead? But I guess we could always change that in the future too, if needed.

psavery:
> fabio.open( fabio.test.utilstest.UtilsTest.getimage("dark_before_000428.edf.ge5.bz2") ).nframes

I just fixed this.

jonwright (approved):
Looks good to me

psavery:
Thanks @jonwright and @kif!

I guess we'll need to wait for a new release on pypi to get access to these features in a release?

kif:
indeed ... I am making 2 to 3 releases a year.
The support for numpy2 is something important, and once h5py is compatible it would make sense to have such a release.

PR #559, opened 2024-03-24
fix documentation build

kif:
Spotted by edna-site's CI


kif:
Not checked by CI ... so bypass tests

PR #563, opened 2024-04-10
CI: Added github action release workflow

t20100:
This PR adds a release workflow that is triggered when publishing a release in github and that can be run manually.

This workflow:
- Builds and tests the source tarball
- Builds and tests the wheels with cibuildwheel
- Builds the documentation and saves it as a zip file
- Publishes the tarball and wheels in pypi when the workflow is triggered by a published release (and after validation)

To make it work it requires to:
- [x] Create an environment in github
- [x] [Add a publisher to pypi](https://docs.pypi.org/trusted-publishers/adding-a-publisher/)

This is adapted from cibuildwheel documentation:
- [Configure a CI service](https://cibuildwheel.pypa.io/en/stable/setup/#github-actions)
- [Deliver to pypi](https://cibuildwheel.pypa.io/en/stable/deliver-to-pypi/#github-actions)

It's similar to [silx's release.yml](https://github.com/silx-kit/silx/blob/main/.github/workflows/release.yml) and [hdf5plugin's release.yml](https://github.com/silx-kit/hdf5plugin/blob/main/.github/workflows/release.yml)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,135 @@
+name: Build and deploy
+
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - published
+
+jobs:
+  build_sdist:
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: "pip"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install build twine
+      - name: Build sdist
+        run: python -m build --sdist
+      - name: Check the package
+        run: python -m twine check dist/*
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cibw-sdist
+          path: dist/*.tar.gz
+
+  test_sdist:
+    needs: [build_sdist]
+    name: Test source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: "pip"
+      - uses: actions/download-artifact@v4
+        with:
+          name: cibw-sdist
+          path: dist
+      - name: Install from sdist
+        run: pip install "$(ls dist/fabio-*.tar.gz)"
+      - name: Run tests
+        run: python -c "import fabio.test, sys; sys.exit(fabio.test.run_tests())"
+
+  build_doc:
+    name: Build documentation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: "pip"
+      - name: Install fabio
+        run: |
+          pip install .
+      - name: Build doc
+        run: |
+          export FABIO_VERSION="$(python -c 'import fabio; print(fabio.strictversion)')"
+          sphinx-build doc/source/ "fabio-${FABIO_VERSION}_documentation/"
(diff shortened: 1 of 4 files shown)


t20100:
- Added `nbsphinx` to `requirements.txt`
- Updated sphinx config to accomodate with [readthedocs changes](https://github.com/readthedocs/readthedocs.org/pull/10638) (and ease documentation generation in release workflow)

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+          cache: "pip"
+      - name: Install pandoc&graphviz
+        run: sudo apt-get install pandoc graphviz
+      - name: Install fabio
+        run: pip install .
+      - name: Install documentation dependencies
+        run: pip install -r requirements.txt
+      - name: Build doc
+        env:
+          READTHEDOCS: "True"  # To skip checking that fabio is installed locally
```
t20100, line 68: This could be remove if the doc's `conf.py` was not checking for place where `fabio` is installed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+          - os: ubuntu-20.04
+            cibw_archs: "ppc64le"
+          - os: windows-2019
+            cibw_archs: "auto64"
+          - os: macos-11
+            cibw_archs: "x86_64"
+            macos_target: "10.9"
+          - os: macos-14
+            cibw_archs: "arm64"
+            macos_target: "11.0"
```
t20100, line 99: Use macos M1 machines.

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+        with:
+          name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
+          path: ./wheelhouse/*.whl
+
+  pypi-publish:
+    needs: [build_doc, build_sdist, build_wheels, test_sdist]
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
```
t20100, line 134: This environment name needs to be declared in pypi.org

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+    permissions:
+      id-token: write
+    if: github.event_name == 'release' && github.event.action == 'published'
```
t20100, line 137: Run only when triggered by a published release

doc/source/conf.py
```diff
@@ -26,7 +23,7 @@
     import fabio
     project_dir = os.path.abspath(os.path.join(__file__, "..", "..", ".."))
     build_dir = os.path.abspath(fabio.__file__)
-    if on_rtd:
+    if os.environ.get('READTHEDOCS') == 'True':
         print("On Read The Docs")
         print("build_dir", build_dir)
         print("project_dir", project_dir)
```
t20100, line 29: Is the whole try/except block needed.
One need to make sure to install the same version of `fabio` as the one that is used to build the doc, but is it worth preventing generating the doc from an installed version?

doc/source/conf.py
```diff
@@ -55,19 +52,11 @@
...
               ]
 
-# Set the theme to sphinx_rtd_theme when *not* building on Read The Docs.
-# The theme is set to default otherwise as Read The Docs uses its own theme anyway.
-if not on_rtd:
-    try:
-        import sphinx_rtd_theme
-        extensions.append('sphinx_rtd_theme')
-    except:
-        print("sphinx_rtd_theme is not available")
```
t20100, line 69: readthedocs no longer set the theme automatically

t20100:
For an example of workflow run, see: https://github.com/t20100/fabio/actions/runs/8633327052 (Artifacts can be downloaded at the bottom of the web page)

Note: The tests of the sdist failed on a test with randomly generated data.

kif:
Let's give it a  try to build a release

t20100:
This run https://github.com/t20100/fabio/actions/runs/8648448087 provides zip with generated wheels for testing

PR #574, opened 2024-07-02
removes _version

### LINKED ISSUE #573: Different import version sources
Seems like there are 2 different ways to get the version.

`from .._version import version as fabio_version` seems to be problematic.

Found here:
https://github.com/silx-kit/fabio/blob/66d472199a8d09ea23c6fd5a661f850833be5aa7/src/fabio/app/eiger2crysalis.py#L58
https://github.com/silx-kit/fabio/blob/66d472199a8d09ea23c6fd5a661f850833be5aa7/src/fabio/app/densify.py#L57

Different call:
https://github.com/silx-kit/fabio/blob/66d472199a8d09ea23c6fd5a661f850833be5aa7/src/fabio/app/eiger2cbf.py#L52

poautran:
close #573

### DIFF
--- a/src/fabio/app/densify.py
+++ b/src/fabio/app/densify.py
@@ -54,7 +54,7 @@ import json
 import numpy
 from .. import eigerimage, limaimage, sparseimage
 from ..openimage import openimage as fabio_open
-from .._version import version as fabio_version
+from .. import version as fabio_version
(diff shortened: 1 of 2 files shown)


kif (approved):
LGTM, I will provide some non regression tests

PR #575, opened 2024-07-02
test all imports

kif:
Remove all `setup.py` and prevent any regression like the one found in #573


kif:
Let's see if it passes under windows ...

kif:
@poautran I believe all tests are now passing on all platforms

PR #580, opened 2024-08-20
Fix `fast_read_data` when reading the last line

### LINKED ISSUE #578: EDFImage.fast_read_data: cannot read a line ?
For a specific use case I want to read a specific column or a specific line of an EDF file.

[`EDFImage.fast_read_roi`](https://github.com/silx-kit/fabio/blob/main/src/fabio/edfimage.py#L1337) seems to have the API for it.

In my use case it works for column but I cannot make it work for lines.

But looking at the implementation it looks like this use case might not be handled. For example [L1358](https://github.com/silx-kit/fabio/blob/main/src/fabio/edfimage.py#L1358) and [L 1370](https://github.com/silx-kit/fabio/blob/main/src/fabio/edfimage.py#L1370) it looks like reading is handled over a specific dimension.

Can you confirm or provide me an example on how to call it for lines ?
Else do you think the use case of 'reading specific line(s)' is relevant ? Or would it be as fast to read the full file directly and subset the line of interest ?

Many thanks,

kif:
close #578

Nice catch by the way !

### DIFF
--- a/src/fabio/edfimage.py
+++ b/src/fabio/edfimage.py
@@ -49,7 +49,7 @@ __authors__ = ["Jérôme Kieffer", "Jon Wright", "Henning O. Sorensen", "Erik Kn
 __contact__ = "jerome.kieffer@esrf.fr"
 __license__ = "MIT"
 __copyright__ = "ESRF"
-__date__ = "01/06/2022"
+__date__ = "20/08/2024"
 
 import os
 import re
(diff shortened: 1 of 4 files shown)


payno (approved):
:heart:  PR with more test than 'source code' :smiley:

kif:
Non regression test are important, especially when one missed a corner case like here.

PR #594, opened 2024-10-07
Convert Eiger-master files to CBF

### LINKED ISSUE #593: `eiger2cbf` does not support Eiger files from Dectris

kif:
close #593

### DIFF
--- a/src/fabio/app/eiger2cbf.py
+++ b/src/fabio/app/eiger2cbf.py
@@ -37,7 +37,7 @@ to CBF and mimic the header from Dectris Pilatus.
 __author__ = "Jerome Kieffer"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
 __licence__ = "MIT"
-__date__ = "03/12/2021"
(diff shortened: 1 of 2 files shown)


woutdenolf:
Thanks for the quick support! I ran this on one of Matthew's master files and it seems to work. I noticed however that the virtual dataset in the master file only links the first file. eiger2cbf cannot take multiple file right? Just a question. Fixing the master file should not be a problem.

woutdenolf (approved):

kif:
IIRC, Dectris's generated master file uses (several) external links rather than one virtual dataset.
That said, taking several input files it rather complicated since the output option should be modified accordingly ...

woutdenolf:
Ah indeed. It contains an external link to the first file

image

I can add the missing external links since `EigerImage` can handle multiple "data*" datasets (sorted by string): https://github.com/kif/fabio/blob/3df93718fcffcd6c35dd41a317453bf9e34dda28/fabio/eigerimage.py#L134.

PR #601, opened 2025-02-07
Fix wavelength

### LINKED ISSUE #600: Default wavelength is not updated for eiger2crysalis
In the xcaliburimage class there is a line for "crystallograhy wavelength" that needs to be an f-string to get it updated...

kif:
close #600


jonwright (approved):
Looks good to me - thanks!

PR #614, opened 2025-08-27
Fix typo: numpy.atleat_2d → numpy.atleast_2d; add non-regression test

### LINKED ISSUE #613: Possible typo
typo in fabio/eigerimage.py on line 189, which reads numpy.atleat_2d rather than numpy.atleast_2d.

--> provide non regression test ?

Spotted by Christian Storm

ramlah7:
## Summary
This PR fixes a small typo in `src/fabio/eigerimage.py` where `numpy.atleat_2d`
was used instead of `numpy.atleast_2d`.  
In addition, it introduces a **non-regression test** to ensure the typo cannot
be reintroduced in the future.

## Changes
- Corrected the typo: `numpy.atleat_2d` → `numpy.atleast_2d`
- Added `tests/test_eigerimage_typo.py`:
  - This test directly inspects `src/fabio/eigerimage.py` as plain text
  - It asserts that the incorrect string `"atleat_2d"` is not present
  - This avoids the need to import the `fabio` package (which requires a build)

## Why
- The typo would cause an `AttributeError` if that code path was executed.
- Adding a test makes sure future contributors don’t accidentally reintroduce
  the misspelling.

## How to run the test
After cloning the repo and installing `pytest`:

```bash
pip install pytest
pytest -q tests/test_eigerimage_typo.py
Expected output:

1 passed in 0.05s
```
Issue Reference
Fixes #613

### DIFF
--- a/tests/test_eigerimage_typo.py
+++ b/tests/test_eigerimage_typo.py
@@ -1,0 +1,8 @@
+from pathlib import Path
+
+def test_no_atleat_typo_in_source():
+    # Locate the source file directly
+    src_file = Path("src/fabio/eigerimage.py")
+    assert src_file.exists(), f"Source file not found: {src_file}"
+    content = src_file.read_text(encoding="utf-8")
+    assert "atleat_2d" not in content, "Typo 'atleat_2d' still present in eigerimage.py"


kif:
Hi RAMLAH,

Thanks for the PR and the review.
The fix looks OK but tests from fabio are still based on unittest (and not pytest, thus no autodiscovery). Thus the test will probably not be integrated into the testsuite.

PR #641, opened 2026-06-06
deprecation: setting an ndarray shape is deprecated since numpy 2.5.0

### LINKED ISSUE #640: numpy 2.4.6 -> 2.5.0rc1: setting array shape is deprecated
https://github.com/silx-kit/fabio/blob/d0054724272b6926bea9740acdd0c7ec52ca8fb8/src/fabio/edfimage.py#L900
```
DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
```
Should change to numpy.reshape to avoid breaking CI tests

woutdenolf:
For reference: https://github.com/silx-kit/silx/issues/4602#issuecomment-4637994839

Needed for the next silx release with https://github.com/silx-kit/silx/pull/4604.

### DIFF
--- a/src/fabio/GEimage.py
+++ b/src/fabio/GEimage.py
@@ -343,7 +343,7 @@ class GeImage(FabioImage):
         data = numpy.frombuffer(raw, datatype).copy()
         if not numpy.little_endian:
             data.byteswap(True)
-        data.shape = rows, cols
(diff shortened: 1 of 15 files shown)


kif (approved):
LGTM

Thanks Wout

PR #642, opened 2026-06-15
Helper function to increase ulimit on linux systems

### LINKED ISSUE #568: Write helper function to unlock the number of opened files ....
With the size of the experiment increasing, it is common to have 1000000 images in 1000 files ... 
thus one encounters a problem with the number of opened files which is limited, at least under unix.

One (half) solution is to set `ulimit` unlimited as done here :

https://github.com/silx-kit/silx/blob/755fa6d3cbe35e1078f65557f6e9e8352b826da5/src/silx/app/view/main.py#L106-L118

It would be nice to have one helper function that would be available in fabio for doing this.

kif:
close #568

### DIFF
--- a/src/fabio/test/test_io_limits.py
+++ b/src/fabio/test/test_io_limits.py
@@ -88,7 +88,7 @@ class TestCli(unittest.TestCase):
         fabio.utils.cli.relax_ulimit()
         try:
             import resource
-        except:
+        except Exception:
             return
         try:
             soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
--- a/src/fabio/test/test_io_limits.py
+++ b/src/fabio/test/test_io_limits.py
@@ -82,11 +82,26 @@ class TestTiff(_CommonIOLimitTest):
     HEADERSIZE = 196
 
 
+class TestCli(unittest.TestCase):
+    def test_ulimit(self):
+        import fabio.utils.cli
+        fabio.utils.cli.relax_ulimit()
+        try:
+            import resource
+        except:
+            return
+        try:
+            soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+        except Exception as err:
+            logger.error(f"{type(err)}: {err}")
+        self.assertEqual(soft, hard, "soft limits have been increased to hard ones")
+
 def suite():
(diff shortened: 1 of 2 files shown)


src/fabio/utils/cli.py
```diff
@@ -163,3 +167,26 @@ def update(self, value, message="", max_value=None):
...
+
+    The (ugly) solution implemented here just increases the
+    soft limit to the hard limit.
+    """
+    try:
+        import resource
+    except ImportError:
+        _logger.debug("No resource module available")
+    else:
+        if hasattr(resource, "RLIMIT_NOFILE"):
```
t20100: import can be done at the top of the module + 

```
    if resource is None:
        _logger.debug("No resource module available")
        return

    if hasattr(resource, "RLIMIT_NOFILE"):
```

src/fabio/test/test_io_limits.py
```diff
@@ -82,11 +82,26 @@ class TestTiff(_CommonIOLimitTest):
...
 
 
+class TestCli(unittest.TestCase):
+    def test_ulimit(self):
+        import fabio.utils.cli
+        fabio.utils.cli.relax_ulimit()
+        try:
+            import resource
+        except Exception:
+            return
```
t20100, line 92: same why not putting imports at the top of the module?

t20100 (approved):
There's a similar one in silx: https://github.com/silx-kit/silx/pull/4613

src/fabio/utils/cli.py
payno, line 187: Under which condition `RLIMIT_NOFILE` does not exists ?

t20100: strictly speaking it is conditionally available:

https://github.com/python/cpython/blob/11f032f904c8019b332a3c367f335e05cde63628/Modules/resource.c#L437-L439

t20100:
Maybe best to avoid duplication with https://github.com/silx-kit/silx/pull/4613, to be discussed.

I propose to move the new helper in a dedicated module to advertise it better that in `cli`.

src/fabio/utils/cli.py
kif: IIRC it depends on the OS: linux OK, MacOS ? FreeBSD ? Hurd?... windows:KO

PR #644, opened 2026-06-15
Fix numpy deprecation

kif:
Fix deprecatoin warning originating from:
- Numpy 2.5
- Python 3.14

### DIFF
--- a/src/fabio/utils/ExternalResources.py
+++ b/src/fabio/utils/ExternalResources.py
@@ -170,7 +170,7 @@ class ExternalResources:
         :param: relative name of the file.
         :return: full path of the locally saved file.
         """
-        logger.debug("ExternalResources.getfile('%s')", filename)
+        logger.debug(f"ExternalResources.getfile('{filename}')")
 
         if not self._initialized:
             self._initialize_data()
@@ -210,9 +210,7 @@ class ExternalResources:
                 return self.getfile(filename)
         else:
             logger.debug(
-                "Trying to download file %s, timeout set to %ss",
-                filename,
-                self.timeout,
+                f"Trying to download file {filename}, timeout set to {self.timeout}s",
             )
             dictProxies = {}
             if "http_proxy" in os.environ:
@@ -226,7 +224,7 @@ class ExternalResources:
             else:
                 opener = urllib.request.urlopen
 
-            logger.debug("wget %s/%s", self.url_base, filename)
+            logger.debug(f"wget {self.url_base}/{filename}")
             try:
                 data = opener(
                     f"{self.url_base}/{filename}", data=None, timeout=self.timeout
@@ -258,12 +256,12 @@ class ExternalResources:
                 self.all_data[filename] = self.get_hash(data=data)
                 self.save_json()
             else:
-                raise RuntimeError("""Could not automatically download test files %s!
+                raise RuntimeError(f"""Could not automatically download test files {filename}!
                     If you are behind a firewall, please set both environment variable
                      http_proxy and https_proxy.
                     This even works under windows !
                     Otherwise please try to download the files manually from
-                    %s/%s""" % (filename, self.url_base, filename))
+                    {self.url_base}/{filename}""")
         return fullfilename
 
     get_file = getfile
@@ -381,7 +379,7 @@ class ExternalResources:
                     If you are behind a firewall, please set the environment variable
                     http_proxy.
                     Otherwise please try to download the files manually from
-                    {self.url_base}""" % (self.url_base, filename))
+                    {self.url_base}""")
 
         raw_file_exists = os.path.isfile(fullfilename_raw)
         gz_file_exists = os.path.isfile(fullfilename_gz)
@@ -431,6 +429,6 @@ class ExternalResources:
             imgs = self.all_data.keys()
         res = []
         for fn in imgs:
-            logger.info("Downloading from silx.org: %s", fn)
+            logger.info(f"Downloading from silx.org: {fn}")
             res.append(self.getfile(fn))
         return res
--- a/src/fabio/utils/ExternalResources.py
+++ b/src/fabio/utils/ExternalResources.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2026 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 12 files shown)


kif:
Data are not properly downloaded, decompressed and recompressed !

kif:
@t20100 I copied in this PR the `ExternalResources` from `silx` and found out it was not behaving properly when recompressing files (not registering their hashes):
In fabIO, data are always compressed in bzip2 format, decompressed and recompressed in gzip to spare bandwidth.
I fixed it in this PR, specifically in:
https://github.com/silx-kit/fabio/pull/644/commits/9e1cec1f536a0208c335e799eabfc1b4c7f6bd21

Could you review specifically this commit and would it make sense to extend compression to `zstd` which is a kind of modernized version of `gzip`, much faster but definitely not as performant as bzip2, compression-wise.

src/fabio/utils/ExternalResources.py
```diff
@@ -64,16 +69,24 @@ def __init__(self, project, url_base, env_key=None, timeout=60):
                             The environment variable is optional: in case it is not set,
                             a directory in the temporary folder is used.
         :param timeout: time in seconds before it breaks
+        :param data_home: Directory in which the data will be downloaded
         """
         self.project = project
         self._initialized = False
         self.sem = threading.Semaphore()
-
+        self.hash = hashlib.sha256
```
t20100, line 88: Why making it an attribute instead of just calling `hashlib.sha256`

src/fabio/utils/ExternalResources.py
```diff
@@ -64,16 +69,24 @@ def __init__(self, project, url_base, env_key=None, timeout=60):
...
-        self.all_data = set()
+        self.all_data = {}
         self.timeout = timeout
-        self._data_home = None
+        self._data_home = data_home
+        self.testdata = ""
+        self.lock = None
+
+    @property
+    def lockfile(self):
```
t20100: private ?
```suggestion
    def _lockfile(self):
```

src/fabio/utils/ExternalResources.py
```diff
@@ -154,169 +218,199 @@ def getfile(self, filename):
...
             except urllib.error.URLError:
                 raise unittest.SkipTest("network unreachable.")
 
-            if not os.path.isdir(os.path.dirname(fullfilename)):
-                # Create sub-directory if needed
-                os.makedirs(os.path.dirname(fullfilename))
+            dirname = os.path.dirname(fullfilename)
+            if not os.path.isdir(dirname):
+                """Create sub-directory if needed"""
+                os.makedirs(dirname)
```
t20100:
```suggestion
            os.makedirs(dirname, exist_ok=True)
```

src/fabio/utils/ExternalResources.py
```diff
@@ -154,169 +218,199 @@ def getfile(self, filename):
...
-                raise RuntimeError(
-                    ("Could not automatically download test images %s!" % filename)
-                    + "If you are behind a firewall, please set both environment variable http_proxy and https_proxy."
-                    + "This works even under windows !"
-                    + "Otherwise please try to download the images manually from"
-                    + "%s/%s" % (self.url_base, filename)
-                )
+                with self.lock:
+                    with open(fullfilename, mode="wb") as outfile:
+                        outfile.write(data)
```
t20100:
```suggestion
                with self.lock, open(fullfilename, mode="wb") as outfile:
                    outfile.write(data)
```

src/fabio/utils/ExternalResources.py
```diff
@@ -154,169 +218,199 @@ def getfile(self, filename):
...
+                    %s""" % (self.url_base, filename))
+
+        try:
+            import bz2
+        except ImportError:
+            raise RuntimeError("bz2 library is needed to decompress data")
+        try:
+            import gzip
+        except ImportError:
+            gzip = None
```
t20100: why not putting imports at the top?

I'm disappointed by "optional module" in the standard lib...

src/fabio/utils/ExternalResources.py
```diff
@@ -98,22 +111,42 @@ def data_home(self):
                 else:
                     name = "uid" + str(os.getuid())
 
-            basename = "%s_testdata_%s" % (self.project, name)
+            basename = f"{self.project}_testdata_{name}"
             data_home = os.path.join(tempfile.gettempdir(), basename)
         if not os.path.exists(data_home):
-            os.makedirs(data_home)
+            try:
+                os.makedirs(data_home)
```
t20100: can be `os.makedirs(data_home, exist_ok=True)`

src/fabio/utils/ExternalResources.py
```diff
@@ -154,169 +218,199 @@ def getfile(self, filename):
...
+                fd.extractall(output)
+                return [os.path.join(output, i) for i in fd.getnames()]
+
+        if lodn.endswith("zip"):
+            with zipfile.ZipFile(full_path, mode="r") as fd:
+                fd.extractall(output)
+                return [os.path.join(output, i) for i in fd.namelist()]
+
+        raise RuntimeError(
+            "Unsupported archive format. Only tar and zip " "are currently supported"
```
t20100:
```suggestion
            "Unsupported archive format. Only tar and zip are currently supported"
```

src/fabio/utils/ExternalResources.py
```diff
@@ -154,169 +218,199 @@ def getfile(self, filename):
...
-            image_list.sort()
-            try:
-                with open(self.testdata, "w") as fp:
-                    json.dump(image_list, fp, indent=4)
-            except IOError:
-                logger.debug("Unable to save JSON list")
-        baseimage = os.path.basename(filename)
-        logger.info("UtilsTest.getimage('%s')" % baseimage)
+
+        if filename in self.all_data:
```
t20100: why changing the `if` condition?

src/fabio/utils/ExternalResources.py
kif: `hashlib.sha256` is constructor, or probably a factory.
In this way, one ensures the hash function used is always the same for all files.

src/fabio/utils/ExternalResources.py
kif: There are historical reasons for that ... if one compiles python manually (historically done at ESRF...) without the libz or libbz2 installed, the build of those python modules is skipped and those modules were not available as part of the standard library.

+1 for putting those modules at the top of the file (but not to remove the try/except.

src/fabio/utils/ExternalResources.py
t20100: why would it be different?

kif: One could want to change the hash function between instances. I find it interesting to have it defined at a single place in the constructor, hard coded (i.e. not modifiable at runtime without ugly monkey-patching) but in a specific place, i.e.not at the place where this factory is actually used.

src/fabio/utils/ExternalResources.py
```diff
@@ -115,10 +126,10 @@ def data_home(self):
             data_home = os.path.join(tempfile.gettempdir(), basename)
         if not os.path.exists(data_home):
             try:
-                os.makedirs(data_home)
-            except OSError as exc:
+                os.makedirs(data_home, exist_ok=True)
```
t20100, line 129: with `exist_ok` you can spare the `if not os.path.exists(data_home):`, but OK as it is

src/fabio/utils/ExternalResources.py
```diff
@@ -251,14 +266,18 @@ def getfile(self, filename):
                     %s/%s""" % (filename, self.url_base, filename))
         return fullfilename
 
+    get_file = getfile
```
t20100, line 267: Is this needed? it's an internal module for fabio, right? so you can break the API if needed

src/fabio/utils/ExternalResources.py
```diff
@@ -312,12 +331,14 @@ def getdir(self, dirname):
                 return [os.path.join(output, i) for i in fd.namelist()]
 
         raise RuntimeError(
-            "Unsupported archive format. Only tar and zip " "are currently supported"
+            "Unsupported archive format. Only tar and zip are currently supported"
         )
 
-    def get_file_and_repack(self, filename):
+    get_dir = getdir
```
t20100, line 335: same, is it needed?

t20100 (approved):

src/fabio/utils/ExternalResources.py
kif: I would not break APIs that easily ... at least not without a deprecation warning.

src/fabio/utils/ExternalResources.py
kif: If the aim is to have the same file into silx and one wants a consistent API, it makes sense to have "_".

PR #645, opened 2026-06-15
Ensure data can be imported from a byte-stream (with some limitations ...)

### LINKED ISSUE #612: Bug: Using `fabio._openimage` with an IO buffer.
## The Problem:

If I use a ByteIO buffer to serve fabio at TIF file (i.e. "marccd/tif"), then the [`fabio.openimage._openimage`](https://github.com/silx-kit/fabio/blob/main/src/fabio/openimage.py#L238) function uses `actual_filename` as the buffer, and the original buffer as `filename` when serving to [`do_magic`](https://github.com/silx-kit/fabio/blob/main/src/fabio/openimage.py#L266C9-L266C18) (which uses the header bytes of the file to decypher the filetype). 

`do_magic` then fails at a [`filename.split`](https://github.com/silx-kit/fabio/blob/main/src/fabio/openimage.py#L152) call, which assumes that filename is only a string. 

## The use case: 
I have files I want to serve and then process over a web server - I'd rather not be required to save them into local storage before pipelining the file stream into fabio.

## Possible solutions:

1. Add another optional argument in `_openimage` for a buffer, then filename can be used if a buffer is provided for a string name. Ideally we modify `openimage` to be able to handle buffers too.
2. Looks to me like all the python interfaces for buffers have a [`name` attribute](https://github.com/python/typeshed/blob/main/stdlib/_io.pyi). However this looks pretty precarious and I don't believe is documented on the [io documentation](https://docs.python.org/3/library/io.html#io.IOBase), but this could be used i.e.: 

        if hasattr(filename, "seek") and hasattr(filename, "read"):
                # Data stream without filename
                filename.seek(0)
                data = filename.read()
                actual_filename = BytesIO(data)
                if hasattr(filename, name) and filename.name:
                    filename = filename.name
                # Back to the location before the read
                filename.seek(0)

3. (easiest?) Modify the logic in `do_magic` when considering `elif format_type == "marccd/tif":`, to not assume filename is string.
    
I'm happy to have a go at contributing if you don't mind and think any fix is worthwhile - keen to contribute more to open source scientific projects! :)

kif:
close #612


woutdenolf (approved):

src/fabio/openimage.py
```diff
@@ -284,7 +291,8 @@ def _openimage(filename):
 
     filetype = None
     try:
-        filetype = do_magic(magic_bytes, filename)
+        filetype = do_magic(magic_bytes,
+                        None if stream_input else filename)
```
woutdenolf: `do_magic` cannot always handle `None`

https://github.com/silx-kit/fabio/blob/f1f577e174a766db389f35108037f5b20eba8914/src/fabio/openimage.py#L131

and I guess this line was to handle the case we had before (filename is an IO stream)

https://github.com/silx-kit/fabio/blob/f1f577e174a766db389f35108037f5b20eba8914/src/fabio/openimage.py#L171

Very tricky stuff.

woutdenolf (commented):
Type annotations really do come in handy to judge this kind of stuff.

I would expect this because `filename` is treated as a string

```python
def do_magic(byts: bytes, filename: str) -> str:
    ...
```

except in one line where `isinstance(filename, str)`. Could you clarify? Most likely `do_magic` needs fixes for the filename type. Perhaps we handle `filename: Optional[str] = None`?

`do_magic` is used in only one place but is part of the public API (if we care).

kif:
Very tricky ... very old code which was mostly untouched for almost twenty years (2007-09-28) !
The initial commit is in 6990855227ede78f11853d26c7b9dcf4c36e7555 and it is from Jon. I was not at ESRF at this time.

Probably type anotation would make things simpler but duck-typing was the way to go at that time.

jonwright:
The initial commit looks a bit different for the `do_magic` signature. Git blame locates 4dc9dee.

woutdenolf (approved):
Looks better. Thanks!

src/fabio/openimage.py
```diff
@@ -122,8 +122,13 @@
 ]
 
 
-def do_magic(byts, filename):
-    """Try to interpret the bytes starting the file as a magic number"""
+def _do_magic(byts:bytes, filename:str="") -> str:
+    """Try to interpret the bytes starting the file as a magic number
+
+    :param byts: begining of the file
+    :param filename: name of the file or None in the case of a buffer
```
woutdenolf:
```suggestion
    :param filename: name of the file or empty string in the case of a buffer
```

PR #647, opened 2026-06-16
Implement sexed-dtype to avoid byte-swapping

### LINKED ISSUE #310: Byteswap in all cases where bpp>1 ...
Spotted in edfimage at least, by Peter Boesecke. Probably affects only "complex128" with swapped endianness.

kif:
Lots of modification all over the code ... and in addition, one has access only to little-endian computers.

Close #310


kif:
Hi reviewers, 
For this silx dev week I decided to set `Claude` on an 7-year-old issue, reported at that time by Peter Boesecke.
Claude managed to migrate all project from swapping bytes to directly reading/writing data with an  "endianness aware data-type", well it was nevertheless a day of work. I took me as long to clean up manually the code, there could be some remaining inconsistencies, so than for the review.

woutdenolf:
I see one `byteswap` call left. Less crucial but good to stay consistent.

https://github.com/silx-kit/fabio/blob/d474564da688d435896c8655df35221fa32a144a/src/fabio/test/codecs/test_brukerimage.py#L60-L65

```python
LE_uint16 = np.dtype("uint16").newbyteorder("<")

MYIMAGE = np.full((256, 256), 16, dtype=LE_uint16)
```

src/fabio/compression/compression.py
```diff
@@ -453,9 +434,8 @@ def decPCK(
     :param overflowPix: optional parameters: number of overflowed pixels
     :param version: PCK version 1 or 2
     :param normal_start: position of the normal value section (can be auto-guessed)
-    :param swap_needed: set to True when reading data from a foreign endianness (little on big or big on little)
+    :param byteorder: set to ">" for big=endian or "<" for little-endian data decompression
```
woutdenolf: https://numpy.org/doc/2.1/reference/generated/numpy.dtype.newbyteorder.html

There are more. Perhaps reference numpy's newbyteorder instead.

src/fabio/ext/mar345_IO.pyx
```diff
@@ -134,7 +134,7 @@ def uncompress_pck(bytes raw not None, dim1=None, dim2=None, overflowPix=None, v
     :param overflowPix: optional parameters: number of overflowed pixels
     :param version: PCK version 1 or 2
     :param normal_start: position of the normal value section (can be auto-guessed)
-    :param swap_needed: set to True when reading data from a foreign endianness (little on big or big on little)
+    :param byteorder: set to ">" to decompress big-endian data else "<" for little-endian
```
woutdenolf: Same.

src/fabio/mar345image.py
```diff
@@ -83,7 +84,7 @@ def __init__(self, *args, **kwargs):
         FabioImage.__init__(self, *args, **kwargs)
         self.numhigh = None
         self.numpixels = None
-        self.swap_needed = None
+        self.byteorder = "="
```
woutdenolf: https://numpy.org/doc/2.1/reference/generated/numpy.dtype.newbyteorder.html

- ‘S’ - swap dtype from current to opposite endian
- {‘<’, ‘little’} - little endian
- {‘>’, ‘big’} - big endian
- {‘=’, ‘native’} - native order
- {‘|’, ‘I’} - ignore (no change to byte order)

To keep the same behavior this would be "|" instead of "=" afaiu.

src/fabio/edfimage.py
```diff
@@ -679,9 +647,9 @@ def _unpack(self):
                 rawData = rawData[:expected]
             # PB38k20190607: explicit way: count = get_data_counts(shape)
             count = self.size // self._dtype.itemsize
-            data = numpy.frombuffer(rawData, self._dtype, count).copy().reshape(shape)
-            if self.swap_needed():
-                data.byteswap(True)
+            stype = self.get_stype(self._dtype, self._data_byteorder)
+            print(stype)
```
woutdenolf:
```suggestion
```

src/fabio/edfimage.py
```diff
@@ -1466,12 +1448,11 @@ def fast_read_data(self, filename=None):
             f.seek(frame.start)
             raw = f.read(frame.blobsize)
         try:
-            data = numpy.frombuffer(raw, dtype=self.bytecode).copy()
+            file_endianness = "big" if (numpy.little_endian == bool(frame.swap_needed())) else "little"
```
woutdenolf: Ok I'm getting too old for this 

```python
numpy.little_endian = machine_is_little
swap_needed = machine_endianness != file_endianness

file_is_big = machine_is_little == swap_needed
file_is_big = machine_is_little == (machine_endianness != file_endianness)
file_is_big = file_endianness == BIG

file_endianness = "big" if file_is_big else "little"
```

This would be much simpler:

```python
file_endianness = (
    "little"
    if frame._data_byteorder is ENDIANNESS.LITTLE
    else "big"
)
```

src/fabio/readbytestream.py
```diff
@@ -58,7 +58,7 @@ def readbytestream(
     nbytespp,
     datatype="int",
     signed="n",
-    swap="n",
+    byteorder="<",
```
woutdenolf, line 61: I think this should be "|" if we want to keep the same behavior.

woutdenolf (commented):
I understand why it took you a day. Jeeeezus. I did a pass but I suggest a third person goes through the PR as well.

kif:
Thanks @woutdenolf you spotted some which went through ...

src/fabio/compression/compression.py
kif: To my knowledge, there are only types of byte-order, all the other fall back on any of the two supported.

This is a python adapter, we could add support for the other, but:
-  "|" makes no sense, data-types are all longer than 1 byte 
- "=" is what we want to prevent because we would not have to manage it if we could
-  "S" has neither sense here since we have no starting point.
Actually none of the alternative solution make sense in this function, so I would prefer keeping it simple: <>

OK for correcting the typo

src/fabio/readbytestream.py
kif: Without "reference point" one cannot keep anything !

src/fabio/edfimage.py
```diff
@@ -853,6 +820,29 @@ def iFrame(self):
...
+        if self._data_byteorder is ENDIANNESS.LITTLE and numpy.little_endian:
+            return False
+        elif self._data_byteorder is ENDIANNESS.BIG and numpy.little_endian:
+            return True
+        elif self._data_byteorder is ENDIANNESS.LITTLE and not numpy.little_endian:
+            return True
+        elif self._data_byteorder is ENDIANNESS.BIG and not numpy.little_endian:
+            return False
+        else:
+            logger.warning("Unconsistent endianness !!!")
```
payno, line 833: As byteorder can only be "little" or "big" I think the following would simplify the code:
I think `_data_byteorder` can be `|` (does not matter) or None. not sure the "does not matter" is handled today.

```python
    def swap_needed(self):
        """
        Decide if we need to byteswap
        """
        if self._data_byteorder is ENDIANNESS.LITTLE:
            return not numpy.little_endian:
        elif self._data_byteorder is ENDIANNESS.BIG:
            return numpy.little_endian:
        else:
            logger.warning("Unconsistent endianness !!!")
```

src/fabio/dm3image.py
```diff
@@ -74,6 +75,9 @@
     20: None,
 }
 
+BE_uint16 = numpy.dtype(">u2")
+BE_uint32 = numpy.dtype(">u4")
+BE_int32 = numpy.dtype(">i4")
```
t20100: Best to be consistent with the other dtypes with `newbyteorder`

t20100 (approved):

src/fabio/dm3image.py
kif: I prefer explicit "int64" instead of "i8" which can be wrongly interpreted as "int8" ... and since it is in the core of the module, it is called only when importing the module.

Issue #653, opened 2026-06-18
include patches from debian ...


kif:
So that they are mo more needed

PR #655, opened 2026-06-18
Update modules.rst

kif:
Apply the patch from debian:

https://sources.debian.org/patches/python-fabio/2025.10.0-3/0001-reproducible.patch/


kif:
close #653

PR #656, opened 2026-06-18
Improve import handling for QtPy

kif:
Modified import statements to handle potential ImportError for QtPy.

Close # 654


kif:
@t20100 
I don't especially like this approach, but maybe you have some suggestion to silent-out `ImportError` when importing a module.

One could have QtPy available inside fabio as a single entry-point if this is finally accepted.

kif:
I twisted CI, now tests are passing, this is probably not useful, at least not for this release.

t20100:
If it's only for the `import_all` filter, why not handling this in this test by skipping the files requiring QtPy when `WITH_QT_TEST` is false?

something like:
```python
def test_import_all(self):
    import fabio

    skip_without_qt = {
        "fabio.app.viewer",
        "fabio.qt.dialogs",
        "fabio.qt.matplotlib",
    }
    with_qt = os.environ.get("WITH_QT_TEST", "True").lower() not in ("0", "false", "no")

    base = os.path.split(fabio.__path__[0])[0] + "/"
    for root, dirs, files in os.walk(fabio.__path__[0]):
        for f in files:
            if f.endswith(".py"):
                module = os.path.join(root, f[:-3])[len(base):].replace(os.sep, ".")
                if not with_qt and module in skip_without_qt:
                    continue
                __import__(module)
```

PR #666, opened 2026-07-28
Two small fixes

loichuder:
- [Fix format missing in relax_ulimit log](https://github.com/silx-kit/fabio/commit/a056c7e7447f683be74ce66c0c6660e033f64fd3)
- [Remove unneeded dep tomli](https://github.com/silx-kit/fabio/commit/12c55230c14e8d1efaf0d393144b70446a9d971a): no longer needed since the min supported version is Python 3.11

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -5,42 +5,41 @@ name: Python package
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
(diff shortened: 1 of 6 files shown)


loichuder:
CI failure due to new `ruff` version that enables more rules: https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#0160

I'll pin `ruff` to `<0.16.0` for now

t20100 (approved):

PR #674, opened 2026-09-17
adds 2D dataset handling to EigerImage reader

skordaschristofanis:
At GSECARS, we collect diffraction data during pulsed heating with the Eiger2 9M detector's external gated-triggering (HDR) mode. As a result, we get an HDF5 file with only 2 dimensions instead of 3.

EigerImage.read() assumes datasets are always 3D (frames * H * W), causing a ValueError when opening HDF5 files where the data is stored as a single 2D image (H * W) with no frame.

### Changes
- adds 2D dataset handling in read(): reads [:, :] directly instead of [frame, :, :]
- adds 2D dataset handling in getframe(): returns the dataset directly instead of None for 2D datasets in multi dataset list
- adds test_read_2d


kif:
Would you have such frame to share ... to make an actual test ?

Beside this, it looks OK to me. Thanks for the contribution

kif:
```
======================================================================
ERROR: test_read_2d (fabio.test.codecs.test_eigerimage.TestEiger.test_read_2d)
check we can read a 2D (single image, no frame axis) Eiger file
----------------------------------------------------------------------
  File "C:\projects\fabio\build\Lib\site-packages\fabio\test\codecs\test_eigerimage.py", line 142, in test_read_2d
    os.unlink(fn)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\fabio_cqjwxr2t_appveyor\\eiger2d.h5'
```

This is likely specific to windows ... file not closed before removing.

skordaschristofanis:
Please see the following zipped .h5 file. The shape is (3262, 3108).

[CC105-DH_0001.zip](https://github.com/user-attachments/files/32363176/CC105-DH_0001.zip)

I added e.close() to handle the windows permissions issue. Thanks.

kif:
Thanks for the file. It can be retrieved via:
`filename = UtilsTest.getimage("CC105-DH_0001.h5.bz2")[:-4]`

skordaschristofanis:
Thanks. I updated the test to use it.

kif (approved):
Looks good to me

kif:
Thanks for the contribution

